PR #8, opened 2021-08-09
Fix missing renames in GitHub Actions.

lmwnshn:
In #7, I forgot to rename the CI databases.


lmwnshn:
Note that this appears to fail in CI because only the version of maven.yml already in `main` will be run.

Issue #33, opened 2021-09-18
Auctionmark shows many SQLExceptions!


amingit21:
Hi there,
I recently installed benchbase (as hoping to get rid of some errors of Oltpbench). but unfortunetely, I still get errors, but of course different than before. In this issue, I focus on Auctionmark on MySQL DB. I set the scale-factor to 100 to make around 25GB of table instead of 260MB. The creation phase goes OK. but in the execution phase, it shows many WARNINGs of SQLException which does not look good.
**Here is one of the output messages:**
```
[WARN ] 2021-09-15 00:31:25,379 [AuctionMarkWorker<000>]  com.oltpbenchmark.api.Worker doWork - SQLException occurred during [com.oltpbenchmark.benchmarks.auctionmark.procedures.NewItem/07] and will not be retried... sql state [23000], error code [1062].
java.sql.SQLIntegrityConstraintViolationException: Duplicate entry '316659366627087-17827599' for key 'item.PRIMARY'
        at com.mysql.cj.jdbc.exceptions.SQLError.createSQLException(SQLError.java:117)
        at com.mysql.cj.jdbc.exceptions.SQLExceptionsMapping.translateException(SQLExceptionsMapping.java:122)
        at com.mysql.cj.jdbc.ClientPreparedStatement.executeInternal(ClientPreparedStatement.java:953)
        at com.mysql.cj.jdbc.ClientPreparedStatement.executeUpdateInternal(ClientPreparedStatement.java:1092)
        at com.mysql.cj.jdbc.ClientPreparedStatement.executeUpdateInternal(ClientPreparedStatement.java:1040)
        at com.mysql.cj.jdbc.ClientPreparedStatement.executeLargeUpdate(ClientPreparedStatement.java:1350)
        at com.mysql.cj.jdbc.ClientPreparedStatement.executeUpdate(ClientPreparedStatement.java:1025)
        at com.oltpbenchmark.benchmarks.auctionmark.procedures.NewItem.run(NewItem.java:172)
        at com.oltpbenchmark.benchmarks.auctionmark.AuctionMarkWorker.executeNewItem(AuctionMarkWorker.java:748)
        at com.oltpbenchmark.benchmarks.auctionmark.AuctionMarkWorker.executeWork(AuctionMarkWorker.java:371)
        at com.oltpbenchmark.api.Worker.doWork(Worker.java:357)
        at com.oltpbenchmark.api.Worker.run(Worker.java:269)
[...]

PR #53, opened 2021-10-27
Extend bit-packing in AuctionMark's ItemId from int to long.

timveil:
fixes issue #33 with AuctionMark


timveil:
while trying to test this fix with scale factors above 100 I believe I stumbled on a serious bug in Thread Util.  See #62.  Testing now with larger scale factors

PR #64, opened 2021-10-29
resolving issue 62; bug in ThreadUtil 

### LINKED ISSUE #62: Bug in ThreadUtil causing benchmarks to fail at high scale factor
I've discovered a problem with ThreadUtil that's exposed during high SF runs... my tests were done at 250 SF on the seats benchmark.  What I observed is that not all loading threads would be permitted to finish even though the logic appeared to be written to wait until all loaders had counted down their latches resulting in a rather confusing set of failures.  Here is a snippet of the current logic...

```java
        final long start = System.currentTimeMillis();
        final int num_threads = runnables.size();
        final CountDownLatch latch = new CountDownLatch(num_threads);
        LatchedExceptionHandler handler = new LatchedExceptionHandler(latch);

        if (LOG.isDebugEnabled()) {
            LOG.debug(String.format("Executing %d threads and blocking until they finish", num_threads));
        }
        for (R r : runnables) {
            pool.execute(new LatchRunnable(r, latch, handler));
        }

        pool.shutdown();


        try {
            latch.await();
        } catch (InterruptedException ex) {
            LOG.error("ThreadUtil.run() was interrupted!", ex);
            throw new RuntimeException(ex);
        } finally {
            if (handler.hasError()) {
                String msg = "Failed to execute threads: " + handler.getLastError().getMessage();
                throw new RuntimeException(msg, handler.getLastError());
            }
        }
        if (LOG.isDebugEnabled()) {
            final long stop = System.currentTimeMillis();
            LOG.debug(String.format("Finished executing %d threads [time=%.02fs]",
                    num_threads, (stop - start) / 1000d));
        }
```

The problem is that `pool.shutdown()` is called before waiting for latches to finish.  Perhaps this isn't a problem during small SF runs but as SF increases this causes issues.  As described in the javadoc for `shutdown()`... "This method does not wait for previously submitted tasks to complete execution. Use awaitTermination to do that.".  Thus loader tasks are potentially killed before they are complete.  The fix reorders this logic so that `shutdown` isn't called until all latches are finished.  In addition to this fix, I added some improved logging, wrapped important code in `finally` blocks and removed a lot of confusing rethrows of exceptions that were wrapped in unnecessary `RuntimeException` 's.

timveil:
this PR fixes #62 which was causing problems at large SF


apavlo (approved):
LGTM

PR #66, opened 2021-10-31
Adding tpcc support for Apache Phoenix

chrajeshbabu:
Added DDLs and Dialect changes required for Apache Phoenix with default configurations.


timveil:
seems like the problem of hbase phoenix support would better be solved by adding it as a type to `DatabaseType` and addressing the differences in sql dialect like we do for other databases via `ddl-*.sql` and `dialect-*.sql` where `*` is the database type.  I would not be in favor of accepting this PR as is.

apavlo:
@timveil I agree with you.

chrajeshbabu:
> seems like the problem of hbase phoenix support would better be solved by adding it as a type to `DatabaseType` and addressing the differences in sql dialect like we do for other databases via `ddl-*.sql` and `dialect-*.sql` where `*` is the database type. I would not be in favor of accepting this PR as is.

Thank you for quick response @timveil, am able to define the ddls dialects for Apache Phoenix. Here I have pushed the changes. But only thing blocking to use the tool is INSERT statement creation in SQLUtil#getInsertSQL. Now I have changed to to use the keyword based on the database type.
Please let me know if that's fine.

chrajeshbabu:
@timveil @apavlo  would you please review the new changes if you have time. Thanks.

timveil (commented):
I think these changes look good.  there is a docker folder where we have docker files for easy local testing of mysql, postgres, cockroach and even spanner.  is there a docker based configuration for phoenix that you can include?  Otherwise i think i'm fine with these changes..  @apavlo ?

chrajeshbabu:
@timveil 
Thank you for reviewing and your time.
There is no docker based configuration for Phoenix at this moment. Will work and try to add as part of other ticket.

apavlo (approved):
LGTM

chrajeshbabu:
Thank you for review @apavlo.

chrajeshbabu:
@apavlo @timveil could you please merge it?

Issue #67, opened 2021-11-01
bug threadutil logic when loader has zero sized runnable collection


timveil:
i introduced this bug during the fix for #62

PR #68, opened 2021-11-01
resolving issue introduced by earlier fix in threadutil

timveil:
this PR fixes issue #67


timveil:
Something is funky with Github actions.  The issue is fixed but the actions aren't picking up the changes.  I think it safe to merge this despite the failing checks. @apavlo

PR #69, opened 2021-11-06
adding support for MariaDB

rusher:
At the same time correcting resource path to permit using windows


src/main/java/com/oltpbenchmark/api/StatementDialects.java
```diff
@@ -84,9 +85,9 @@ public String getSQLDialectPath(DatabaseType databaseType) {
 
         if (fileName != null) {
 
-            final String path = "benchmarks" + File.separator + workloadConfiguration.getBenchmarkName() + File.separator + fileName;
+            final String path = "/benchmarks/" + workloadConfiguration.getBenchmarkName() + "/" + fileName;
```
apavlo, line 88: Isn't this going to break on unix platforms?

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
```diff
@@ -189,16 +189,20 @@ public String getDatabaseDDLPath(DatabaseType db_type) {
         // The order matters!
         List<String> names = new ArrayList<>();
         if (db_type != null) {
-            names.add("ddl-" + db_type.name().toLowerCase() + ".sql");
+            if (DatabaseType.MARIADB == db_type) {
+                // common ddl for mariadb and mysql
+                names.add("ddl-mysql.sql");
```
apavlo, line 194: Instead of hardcoding the string "-mysql", you should override the `db_type`:
```
DatabaseType ddl_db_type = db_type;
if (ddl_db_db_type == DatabaseType.MARIADB) ddl_db_type = DatabaseType.MYSQL;
names.add("ddl-" + ddl_db_type.name().toLowerCase() + ".sql");
```

src/main/java/com/oltpbenchmark/util/ScriptRunner.java
```diff
@@ -55,7 +55,7 @@ public void runScript(String path) throws IOException, SQLException {
 
         LOG.debug("trying to find file by path {}", path);
 
-        try (InputStream in = this.getClass().getClassLoader().getResourceAsStream(path);
+        try (InputStream in = this.getClass().getResourceAsStream(path);
```
apavlo, line 58: What are the implications of this change?

apavlo (changes requested):
Thanks for sending this! I have some questions / feedback

src/main/java/com/oltpbenchmark/api/StatementDialects.java
rusher: no, tested it (and you can see the results on unix travis tests [here](https://github.com/rusher/benchbase/actions/runs/1429002060))

src/main/java/com/oltpbenchmark/util/ScriptRunner.java
rusher: I was on a windows machine when doing that PR, and windows resource access works a little bit differently than unix. Using class.getResourceAsStream compare to class.getClassLoader().getResourceAsStream is that path can be relative. I've changed resource path to use absolute path, that way it doesn't change anything, but resolve the small unix/windows quircks.

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
rusher: _corrected_

rusher:
btw, github action for PR run main branch, not the PR commit (because of config `pull_request_target`). I imagine that this might be done to avoid having some secret variable public, but if this is the case, better to completly avoid github action running for PR.

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
```diff
@@ -283,7 +284,9 @@ public final void createDatabase(DatabaseType dbType, Connection conn) throws SQ
                     LOG.debug(String.format("Starting %d %s.LoaderThreads [maxConcurrent=%d]", loaderThreads.size(), loader.getClass().getSimpleName(), maxConcurrent));
                 }
 
-                ThreadUtil.runNewPool(loaderThreads, maxConcurrent);
+                if (!loaderThreads.isEmpty()) {
```
rusher, line 287: this correct #67, but can be removed if #68 is merged

lmwnshn: #68 has been merged.

rusher:
updated with recent commit on main branch (and removed BenchmarkModule.java correction, since #68 is merged)

apavlo (approved):
LGTM

rusher:
any chance to see workflow approval then merge ?

lmwnshn:
@rusher Sorry, this slipped through. Happy holidays!

PR #74, opened 2021-11-11
Merge high scale factor changes

timveil:
I made a series of commits over a period of days working thru testing both SEATS and AuctionMark benchmarks at scale factors of 250.  There were many issues encountered which resulted in significant changes to how CompositeId works, data type/size changes to schema, fixing broken tests, improved logging, etc.  I tested these changes on HSQLDB, Postgres, MySql and Cockroach.  

As it relates to compositeId changes... i recognize their may be a better way to do this, but for now this gets us out of the limitations imposed by bitshifting ints/longs.


timveil:
wrong branch selected... closing as this was an error

PR #75, opened 2021-11-11
Merge series of changes to fix SF250 for AuctionMark and SEATS

timveil:
I made a series of commits over a period of days working thru testing both SEATS and AuctionMark benchmarks at scale factors of 250. There were many issues encountered which resulted in significant changes to how CompositeId works, data type/size changes to schema, fixing broken tests, improved logging, etc. I tested these changes on HSQLDB, Postgres, MySql and Cockroach.

As it relates to compositeId changes... i recognize their may be a better way to do this, but for now this gets us out of the limitations imposed by bitshifting ints/longs.  Since these changes touched many files, i'm happy to talk thru or discuss any/all of the approach


src/main/resources/benchmarks/auctionmark/ddl-generic.sql
```diff
@@ -23,7 +23,7 @@ CREATE TABLE config_profile (
                                 cfp_scale_factor            FLOAT NOT NULL,
                                 cfp_loader_start            TIMESTAMP NOT NULL,
                                 cfp_loader_stop             TIMESTAMP NOT NULL,
-                                cfp_user_item_histogram     VARCHAR(12000) NOT NULL
+                                cfp_user_item_histogram     LONGVARCHAR NOT NULL
```
apavlo, line 26: We should use `TEXT` for the generic schema. That's more common in DBMSs than `LONGVARCHAR`.

apavlo (changes requested):
Minor request to fix auctionmark-generic DDL

src/main/resources/benchmarks/auctionmark/ddl-generic.sql
timveil: so my understanding is that `generic` ddl is used for creating the hsqldb schema.  i took a quick look at the hsqldb docs and didn't see a `STRING` or `TEXT` data type and that `LONGVARCHAR` was the equivalent.  I used `TEXT` for the other flavors of `ddl-*.sql`.

timveil: Hey Andy any more thoughts here?

apavlo: I think LONGVARCHAR is going to break most systems. It looks like we should run HSQLDB in MySQL compatibility mode:

https://stackoverflow.com/a/12290408

I won't have time to test this today.

apavlo (approved):
LGTM

PR #83, opened 2021-12-05
fixes issue #76; tpcc key time and wait time

### LINKED ISSUE #76: implement "keying time" and "think time" option for TPC-C
TPC-C calls for keying time and think time.  would like to add support these wait times in a configurable way so users can choose whether to run TPC-C with or without these times.

timveil:
I added "keying time" and "wait time" configuration options to the `*_config.xml` and applied these changes throughout.  In addition I attempted to clean up the TPCC procedure code.  Much of this code had very large methods that opened and closed many prepared statements inside a single method.  I found these methods very hard to read and troubleshoot.  Much of my changes inside the `tpcc` package were to these procedure classes where I extracted methods to operate on a single prepared statement.  I recognize that comparing these changes will be a bit difficult but I'm confident the behavior should remain the same.  In addition I did some method and variable renaming for clarity.  Examples of this include renaming methods on the `com.oltpbenchmark.util.ResultUploader` and renaming variables to include their time type (`int[] values` to `int[] valuesAsMicroseconds`).  I also cleaned up some logging output and fixed misspellings.  This PR closes #76.


src/main/java/com/oltpbenchmark/util/ResultUploader.java
```diff
@@ -67,16 +69,20 @@ public ResultUploader(Results r, XMLConfiguration conf, CommandLine argsLine) {
 
     }
 
-    public void writeDBParameters(PrintStream os) {
+    public void writeParams(PrintStream os) {
```
apavlo, line 72: AFAIK, nobody is using this feature. The website that handles the uploads hasn't been worked on in eight years. I think we can just delete this class.

apavlo (changes requested):
I think calling the new feature "keyingTime" and "thinkTime" will be confusing. What if we just called them "preExecuteWait" and "postExecuteWait"?

These times should be defined on a per `TransactionType` basis in the config instead of your embedded switch statements.

Then the flags to enable them should be inside of the `<work>` field in the XML config so that we can disable/enable them on a per workload phase.

src/main/java/com/oltpbenchmark/benchmarks/tpcc/TPCCWorker.java
```diff
@@ -74,4 +74,44 @@ protected TransactionStatus executeWork(Connection conn, TransactionType nextTra
...
+    }
+
+    @Override
+    protected long getThinkTimeInMillis(TransactionType type) {
+        // TPC-C 5.2.5.4: For think times for each type of transaction.
+        long mean = switch (type.getName()) {
+            case "NewOrder", "Payment" -> 12000;
+            case "OrderStatus" -> 10000;
+            case "Delivery", "StockLevel" -> 5000;
+            default -> throw new IllegalArgumentException("invalid TransactionType provided: " + type);
```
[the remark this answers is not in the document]
timveil: got it... will do.

src/main/java/com/oltpbenchmark/benchmarks/tpcc/TPCCWorker.java
```diff
@@ -74,4 +74,44 @@ protected TransactionStatus executeWork(Connection conn, TransactionType nextTra
...
+        }
+        if (type.getName().equals("OrderStatus")) {
+            return 2000;
+        }
+        if (type.getName().equals("Delivery")) {
+            return 2000;
+        }
+        if (type.getName().equals("StockLevel")) {
+            return 2000;
+        }
```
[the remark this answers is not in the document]
timveil: got it... will do.

src/main/java/com/oltpbenchmark/util/ResultUploader.java
timveil: So as far as i can tell i already killed all the "upload" features of this class during my refactor.  as of now this class just collects additional data and ultimately writes to various files.  i've been looking at some of these files recently and think they are pretty useful.  again, i think all the "upload" stuff is long gone.

src/main/java/com/oltpbenchmark/DistributionStatistics.java
```diff
@@ -142,16 +142,15 @@ public double getMaximum() {
...
-        return "[min=" + getMinimum() / 1e6 + ", "
-                + "25th=" + get25thPercentile() / 1e6 + ", "
-                + "median=" + getMedian() / 1e6 + ", "
-                + "avg=" + getAverage() / 1e6 + ", "
-                + "75th=" + get75thPercentile() / 1e6 + ", "
-                + "90th=" + get90thPercentile() / 1e6 + ", "
-                + "95th=" + get95thPercentile() / 1e6 + ", "
-                + "99th=" + get99thPercentile() / 1e6 + ", "
-                + "max=" + getMaximum() / 1e6 + "]";
+        return "in microseconds [min=" + getMinimum() + ", "
```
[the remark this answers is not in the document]
timveil: this confused me at first and thus some of the renaming. just to clarify... properties like `getMedian()` return values in microseconds but you would prefer the `toString` method to convert these values to milliseconds by dividing by `1e6`.  Is that correct?

src/main/java/com/oltpbenchmark/benchmarks/tpcc/TPCCLoader.java
```diff
@@ -501,51 +522,124 @@ protected void loadOrders(Connection conn, int w_id, int districtsPerWarehouse,
...
+
+            for (int d = 1; d <= districtsPerWarehouse; d++) {
+
+                for (int c = 1; c <= customersPerDistrict; c++) {
 
                     // 900 rows in the NEW-ORDER table corresponding to the last
                     // 900 rows in the ORDER table for that district (i.e.,
                     // with NO_O_ID between 2,101 and 3,000)
                     if (c >= FIRST_UNPROCESSED_O_ID) {
+                        NewOrder new_order = new NewOrder();
```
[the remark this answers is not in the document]
timveil: maybe we can tackle this in a separate issue/pr.  that would take quite a bit more work

timveil:
> Then the flags to enable them should be inside of the `<work>` field in the XML config so that we can disable/enable them on a per workload phase.

So I did not alter `<work>` and just added the elements as children of `<transactiontype>`.  if either `<preExecutionWait>3000</preExecutionWait>` or `<postExecutionWait>12000</postExecutionWait>` are provided with non-zero values, the wait is applied.  I didn't see a reason to have a separate "enabled" flag.  if i'm missing something, please clarify.

src/main/java/com/oltpbenchmark/util/ResultUploader.java
apavlo: Maybe rename it to `ResultWriter`?

src/main/java/com/oltpbenchmark/benchmarks/tpcc/TPCCLoader.java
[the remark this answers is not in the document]
apavlo: Agreed.

apavlo:
If the official benchmark spec (e.g., TPC-C) says that there should be these wait times, then we should include in the config files. But most people (academics?) are not going to want to execute the workloads with them, so I was thinking we could enable/disable them as needed in the `<work>` portion of the config.

timveil:
> If the official benchmark spec (e.g., TPC-C) says that there should be these wait times, then we should include in the config files. But most people (academics?) are not going to want to execute the workloads with them, so I was thinking we could enable/disable them as needed in the `<work>` portion of the config.

i think i need to go back and do more to understand `<work>`. my changes are breaking TPC-H in part because its the only benchmark with `serial` as `true`.  i'm currently stuck trying to understand what `fetchWork` does.

maybe once i get thru this it will be clearer to me how to enable/disable with `work`

src/main/java/com/oltpbenchmark/util/ResultUploader.java
timveil: done.  also moved "write" methods out of `Results` into newly named `ResultsWriter`

timveil:
@apavlo does `<work>` map to `com.oltpbenchmark.Phase`?  It seems like we parse `<work>` to create an instance of Phase.  If this is the case, i'd like do some refactoring here... if nothing else sync the names for clarity.  Maybe i can tackle enable/disable of waits during that change.   I'd want to do that under a separate issue/PR.  With the exception of those changes I think this PR is good to go.

apavlo:
@timveil Yes, `<work>` is instantiated as the `Phase` class (it is supposed to be `WorkPhase`).

src/main/java/com/oltpbenchmark/benchmarks/tpcc/TPCCWorker.java
```diff
@@ -74,4 +73,25 @@ protected TransactionStatus executeWork(Connection conn, TransactionType nextTra
...
+    @Override
+    protected long getPostExecutionWaitInMillis(TransactionType type) {
+        // TPC-C 5.2.5.4: For think times for each type of transaction.
+        long mean = type.getPostExecutionWait();
+
+        float c = this.getBenchmarkModule().rng().nextFloat();
+        long thinkTime = (long) (-1 * Math.log(c) * mean);
+        if (thinkTime > 10 * mean) {
+            thinkTime = 10 * mean;
+        }
```
apavlo, line 92: Is this logic specific to TPC-C?

src/main/java/com/oltpbenchmark/api/Worker.java
```diff
@@ -182,170 +184,202 @@ synchronized public void cancelStatement() {
...
+
+                if (preExecutionWaitInMillis > 0) {
+                    try {
+                        LOG.debug("{} will sleep for {} ms before executing", transactionType.getName(), preExecutionWaitInMillis);
+
+                        Thread.sleep(preExecutionWaitInMillis);
+                    } catch (InterruptedException e) {
+                        LOG.error("Pre-execution sleep interrupted", e);
                     }
                 }
```
apavlo, line 276: I'm okay with adding this check here for every txn invocation in this PR, but please create a new issue so that we can add the boolean flag to enable/disable.

apavlo (commented):
Should we have benchmark specific wait time calculations? That seems like we are adding more complexity to each benchmark when it could be something globally defined for all benchmarks.

src/main/java/com/oltpbenchmark/benchmarks/tpcc/TPCCWorker.java
timveil: i believe so... 

```
5.2.5.4 For each transaction type, think time is taken independently from a negative exponential distribution. Think time, Tt, is computed from the following equation:
Tt = -log(r) *  where: log = natural log (base e) Tt = think time r = random number uniformly distributed between 0 and 1  = mean think time
Each distribution may be truncated at 10 times its mean value
```

src/main/java/com/oltpbenchmark/api/Worker.java
timveil: will address here... #86

apavlo:
Not sure why Github won't let me update this PR with the latest from main.

timveil:
Want me to merge main into this branch?

apavlo:
@timveil Yes

timveil:
@apavlo `main` has been merged into `issue-76`

src/main/java/com/oltpbenchmark/DistributionStatistics.java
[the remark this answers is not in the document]
timveil: done

apavlo (approved):
LGTM

PR #85, opened 2021-12-05
fixes #55; removing ant, ivy and dockerfile that referenced them

### LINKED ISSUE #55: Remove references to Ivy/Ant from project
Is it possible to remove ivy from this project?  I see that it is used by a `Dockerfile` and specifically its referenced `.deploy` directory.  If this Dockerfile is useful then at least it should be migrated to use Maven instead of Ant.  Seems duplicative to have both ivy and maven listing and managing dependencies.  Thoughts?

timveil:
fixes #55


apavlo (approved):
LGTM

Issue #89, opened 2021-12-13
Smallbank SendPayment transaction has two identical accounts


yqmmm:
https://github.com/cmu-db/benchbase/blob/bbe8c1db84ec81c6cdec6fbeca27b24b1b4e6612/src/main/java/com/oltpbenchmark/benchmarks/smallbank/SmallBankWorker.java#L124-L126

Smallbank SendPayment transaction has two identical accounts.

I understand this transaction did not exists in the original Smallbank benchmark and is added in benchbase. But sending payments from and to the same account does not seem right?

PR #90, opened 2021-12-13
Fix SendPayment account.

yqmmm:
See #89


timveil (commented):
thanks for submitting @yqmmm.  I'm not sure why this file didn't maintain history but looking at the history from OLTPBench this has been part of the codebase/logic since 2016 when it was added by @apavlo... lets see what he has to say about it.

https://github.com/oltpbenchmark/oltpbench/commits/master/src/com/oltpbenchmark/benchmarks/smallbank/SmallBankWorker.java

apavlo:
I think this is a typo from when the SmallBank benchmark code was ported from H-Store to the original OLTP-Bench repo. The original H-Store implementation of the worker/client looks correct:

https://github.com/apavlo/h-store/blob/master/src/benchmarks/edu/brown/benchmark/smallbank/SmallBankClient.java#L101-L105

So yes, this PR fixes the problem.

Thanks @yqmmm!!

PR #111, opened 2022-02-01
Fixed running benchbase.jar from any directory

shuhaowu:
Previously, the current working directory had to be set to the same location as the directory that contains benchbase.jar, as the code expects a `config/plugins.xml` file to exist at the current directory.

This patch changes this so the jar and therefore the config/plugins.xml location is automatically discovered. This allows benchbase to run from any directory by doing `java -jar /path/to/benchbase.jar`.


lmwnshn (approved):
@shuhaowu Thanks for this PR! Apologies for the delay in review.

LGTM (I have tested with TPC-H create+load+exec from root dir).

We will merge this PR, but generally this remains an officially unsupported use-case and we expect people to be in the extracted directory.

shuhaowu:
Thanks! Out of curiosity, is it possible to run with a custom (private) benchmark without needing to rebuild benchbase entirely? If not, is this something benchbase is interested in? We can also discuss this in a separate issue if you'd like.

lmwnshn:
@shuhaowu Please open a separate issue for the discussion, I will bring this up with the people who are helping to maintain benchbase. Thank you!

PR #121, opened 2022-03-03
Resolves Issue 97 - classpath conflicts from various driver jars 

### LINKED ISSUE #97: classpath conflicts due to transitive dependencies of spanner and mysql drivers
both the spanner jdbc driver and the mysql driver specify `com.google.protobuf:protobuf-java` as a dependency.  Mysql version `8.0.27` depends on version `3.11.4` while spanner version `2.5.5` depends on version `3.19.1`.  When running spanner tests, the mysql required version is loaded causing `NoSuchMethod` exceptions at runtime.

timveil:
This PR closes #97.  To accomplish this I introduced the concept of maven profiles and moved database specific drivers from the default dependency list into profile specific dependency configurations.  This means that instead of their being a single build/distribution of benchbase, there is now a build for each supported database.  This eliminates the existing (#97) and future classpath challenges caused by shipping benchbase with every supported driver in the classpath.  

These above changes also required me to make changes to the way github actions executes builds and tests.  I was able to refactor our actions using `matrix` strategies to simplify the configuration and introduce more parallelization. 

The readme file, including build instructions have been updated to reflect this new approach


.github/workflows/maven.yml
```diff
@@ -18,44 +18,57 @@ on:
...
+        name: benchbase-${{matrix.profile}}
+        path: target/benchbase-${{matrix.profile}}.tgz
 
   mariadb:
-    needs: build-and-upload
+    needs: package-and-upload
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        benchmark: [ 'auctionmark', 'epinions', 'hyadapt', 'noop', 'resourcestresser', 'seats', 'sibench', 'smallbank', 'tatp', 'tpcc', 'twitter', 'voter', 'wikipedia', 'ycsb' ]
```
apavlo, line 71: Nice :+1:

apavlo (approved):
LGTM

PR #126, opened 2022-03-08
Switch txns to return instead of unnecessarily aborting in SEATS

### LINKED ISSUE #125: aborted transactions in SEATS
There are a number of instances in SEATS where a `UserAbortException` is thrown when it appears simply escaping the method would do.  This is particularly prevalent in `UpdateReservation`, `DeleteReservation` and `UpdateCustomer`.  This issue is to track the PR that will attempt to reduce the frequency of Aborted Transactions in the SEATS benchmark.

timveil:
this PR resolves #125


src/main/java/com/oltpbenchmark/benchmarks/seats/util/ErrorType.java
```diff
@@ -35,7 +33,6 @@
     UNKNOWN;
 
     private final String errorCode;
-    private final static Pattern p = Pattern.compile("^(USER ABORT:[\\s]+)?E([\\d]{4})");
```
apavlo, line 38: I don't remember what I was using this for?

apavlo (approved):
LGTM

PR #134, opened 2022-03-14
Configuration Refactor

timveil:
This PR contains a number of changes related to a significant refactor of configuration xml files and their associated classes.  The changes largely fall into the following buckets...

- separate existing config file into three separate files... `database`, `workload` and `dialect`.  `database` contains only configuration about the target database.  `workload` contains only information about the benchmark or `workload`.  `dialect` contains only statement level overwrites.
-  each new config file is now validated with a new `.xsd` file
- xpath and other legacy jars/frameworks have been removed in favor of Jackson Xml for mapping of xml documents to Java Objects. 
- the Java representation of the configuration now uses the new `record` java type

The net result of these changes among other things is a simplification of the logic in `DBWorkload`.

In addition to the above changes, another fix was inadvertently added to this PR.  The `wikipedia` benchmark incorrectly threw 100% aborted transactions for 3 procedures.  This issue is resolved as part of this PR


README.md
```diff
@@ -34,10 +34,10 @@ tar xvzf benchbase-postgres.tgz
 cd benchbase-postgres
 ```
 
-Inside this folder, you can run BenchBase. For example, to execute the `tpcc` benchmark,
+Inside this folder, you can run BenchBase. For example, to execute the `tpcc` benchmark (workload) against a `postgres`...
```
lmwnshn, line 37:
```suggestion
Inside this folder, you can run BenchBase. For example, to execute the `tpcc` benchmark (workload) against a `postgres` instance:
```

config/workload/auctionmark.xml
```diff
@@ -0,0 +1,31 @@
...
+
+    <workloads>
+        <workload benchmarkClass="com.oltpbenchmark.benchmarks.auctionmark.AuctionMarkBenchmark">
+            <scaleFactor>1</scaleFactor>
+            <terminals>1</terminals>
+
+            <phases>
+                <phase>
+                    <time>60</time>
+                    <rateType>LIMITED</rateType>
```
lmwnshn, line 12: This seems to be a new parameter; trying to find the documentation through github's file browser (if any), I assume the only other option would be called something like `UNLIMITED` in which case `<rate>` is ignored?

timveil: great catch... this was one of the harder sections of DBWorkload to port over.  There was old logic like this...

```
 } else if (rate_string.equals(RATE_UNLIMITED)) {
                    rateLimited = false;
}...
```

my thinking here was that instead of inferring the ratetype by the presence or absence of UNLIMITED, we ask users to be explicit, therefore this element is always required and the logic becomes clearer.  In the new world a phase can be LIMITED, UNLIMITED or DISABLED.  hopefully this explanation helps and the logic is sound.  Have a look at the new method ` private static List<Phase> getPhases(Workload workload)` on DBWorkload.

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -42,358 +36,79 @@
...
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
 
 public class DBWorkload {
     private static final Logger LOG = LoggerFactory.getLogger(DBWorkload.class);
-
     private static final String SINGLE_LINE = StringUtil.repeat("=", 70);
 
-    private static final String RATE_DISABLED = "disabled";
```
timveil, line 52: @lmwnshn here were the old valued that i converted into the PhaseRateType enum and added the LIMITED value.  see (https://github.com/timveil-cockroach/benchbase/blob/61518dd6ef7ff49ba3c0bf2fec2d26c427184602/src/main/java/com/oltpbenchmark/api/config/PhaseRateType.java#L3)

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -475,35 +207,86 @@ public static void main(String[] args) throws Exception {
         }
     }
 
-    private static Options buildOptions(XMLConfiguration pluginConfig) {
+    private static List<Phase> getPhases(Workload workload) {
```
timveil, line 210: @lmwnshn here is the new method where i attempted to convert the phase logic and the interpret the new enum values

src/test/resources/log4j.properties
```diff
@@ -8,6 +8,7 @@ log4j.appender.A1.layout=org.apache.log4j.PatternLayout
 log4j.appender.A1.layout.ConversionPattern=[%-5p] %d [%t] %x %c %M - %m%n
 
 log4j.logger.com.oltpbenchmark=INFO
+log4j.logger.com.oltpbenchmark.util.ThreadUtil=TRACE
```
lmwnshn, line 11: Do we want this to be part of the defaults? Not sure how verbose that would be.

src/test/java/com/oltpbenchmark/catalog/TestCatalog.java
```diff
@@ -60,8 +60,7 @@
 //        for (Entry<String, String> e : origTableNames.entrySet()) {
 //            assertFalse(e.toString(), e.getKey().equals(e.getValue()));
 //            assertTrue(e.toString(), e.getKey().equalsIgnoreCase(e.getValue()));
-//        } // FOR
-//    }
+//        }//    }
```
lmwnshn, line 63: I guess we used a find and replace regex and it probably should have preserved the line break during the replace; this (and another test file below) pulled the comment upwards, which can potentially break if block-uncommented like in this scenario. But it probably doesn't really matter.

src/test/resources/log4j.properties
timveil: this is the test logging configuration.  ThreadUtil does not output much but what it does print is important to understanding its behavior.  TRACE on this file only adds a few lines to the output.  I'm happy to make a change here but don't see it as a big deal

src/test/java/com/oltpbenchmark/catalog/TestCatalog.java
timveil: hehe.  i'm desperately want to delete all this commented out code.  thats what version control is for :)

src/test/java/com/oltpbenchmark/benchmarks/twitter/TestTwitterLoader.java
```diff
@@ -27,7 +27,6 @@
     @Override
     protected void setUp() throws Exception {
         super.setUp(TwitterBenchmark.class, IGNORED_TABLES, TestTwitterBenchmark.PROC_CLASSES);
-        this.workConf.setScaleFactor(0.1);
```
lmwnshn, line 30: I see this change with other test loaders, does it not cause the scalefactor to be too large?

src/test/java/com/oltpbenchmark/catalog/TestCatalog.java
lmwnshn: I personally have no objections to deleting dead code, if @apavlo doesn't mind.

src/test/java/com/oltpbenchmark/benchmarks/twitter/TestTwitterLoader.java
timveil: so this is addressed by setting a default ScaleFactor in the parent AbstractTestCase instead of relying on individual test loaders to think of it.  see https://github.com/cmu-db/benchbase/blob/76917cf2812c02043cf09458931af0c2f70ccf4b/src/test/java/com/oltpbenchmark/api/AbstractTestCase.java#L52

timveil: so short answer is no.  all test cases will use the smaller scale factor

config/workload/workload.xsd
```diff
@@ -0,0 +1,144 @@
...
+    <xs:complexType name="workloadType">
+        <xs:sequence>
+            <xs:element type="xs:double" name="scaleFactor">
+                <xs:annotation>
+                    <xs:documentation>the scale factor of this workload. this value usually dictates the size of the data generated for this benchmark.</xs:documentation>
+                </xs:annotation>
+            </xs:element>
+            <xs:element type="xs:int" name="terminals">
+                <xs:annotation>
+                    <xs:documentation>the number terminals for this workload. this value usually dictates the number of parallel threads executed during this benchmark.</xs:documentation>
```
lmwnshn, line 104:
```suggestion
                    <xs:documentation>the number of terminals for this workload. this value usually dictates the number of parallel threads executed during this benchmark.</xs:documentation>
```

src/test/java/com/oltpbenchmark/api/TestConfigurationLoad.java
```diff
@@ -0,0 +1,270 @@
...
+                            <statement name="insertItemAttribute">INSERT INTO item_attribute(ia_id,ia_i_id,ia_u_id,ia_gav_id,ia_gag_id) VALUES (?, ?, ?, ?, ?)</statement>
+                            <statement name="deleteItemAttribute">DELETE FROM item_attribute WHERE ia_id = ? AND ia_i_id = ? AND ia_u_id = ?</statement>
+                            <statement name="updateItem">UPDATE item SET i_description = ?, i_updated = ? WHERE i_id = ? AND i_u_id = ?</statement>
+                        </statements>
+                    </procedure>
+                </procedures>
+            </dialect>
+            """;
+
+    public void testLoadDatabase() throws JsonProcessingException {
```
lmwnshn, line 224: If I'm reading this file correctly, all the tests just print the final string at the end. So they're useful if ran and inspected visually, but not useful for CI "did something break" purposes.

Thoughts on saving the string that is currently output and comparing against it to make sure it doesn't change? Admittedly, that would also be a very brittle test. Just thinking out loud

timveil: Good comment and suggestion.  I'll update the test.  I did this quickly as a way to manually check my work but I can make it more robust.

src/main/resources/benchmarks/tpch/ddl-generic.sql
```diff
@@ -1,21 +1,27 @@
...
+DROP TABLE IF EXISTS orders CASCADE;
+DROP TABLE IF EXISTS customer CASCADE;
+DROP TABLE IF EXISTS lineitem CASCADE;
 
 -- Sccsid:     @(#)dss.ddl  2.1.8.1
-CREATE TABLE nation  ( n_nationkey  INTEGER NOT NULL,
-                       n_name       CHAR(25) NOT NULL,
-                       n_regionkey  INTEGER NOT NULL,
-                       n_comment    VARCHAR(152));
+CREATE TABLE nation
```
lmwnshn, line 11: I found it a bit odd that only this part got formatted (as opposed to the entire file)

src/main/resources/benchmarks/seats/ddl-generic.sql
```diff
@@ -1,24 +1,25 @@
...
 -- CONFIG_PROFILE
 --
-CREATE TABLE config_profile (
-                                cfp_scale_factor        float                               NOT NULL,
-                                cfp_aiport_max_customer varchar(10001)                      NOT NULL,
-                                cfp_flight_start        timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
-                                cfp_flight_upcoming     timestamp DEFAULT CURRENT_TIMESTAMP NOT NULL,
-                                cfp_flight_past_days    int                                 NOT NULL,
+CREATE TABLE config_profile
+(
```
lmwnshn, line 17: Similar comment elsewhere, slightly odd to only format the first couple of tables in the .sql files. Not really a problem

src/main/java/com/oltpbenchmark/benchmarks/twitter/TwitterLoader.java
```diff
@@ -287,7 +287,7 @@ protected void loadFollowData(Connection conn, int lo, int hi) throws SQLExcepti
 
                         total++;
                         batchSize++;
-                        f++;
+
```
lmwnshn, line 290: Is this the desired behavior? I don't know; probably @apavlo would know. Now `f` can increase without a new followee.

src/main/java/com/oltpbenchmark/api/config/Transaction.java
```diff
@@ -0,0 +1,7 @@
+package com.oltpbenchmark.api.config;
+
+public record Transaction(Class<? extends com.oltpbenchmark.api.Procedure> procedureClass,
```
lmwnshn, line 3: (unproductive comment) This is the first new Java feature in the wild that I've needed to look up. neat!

src/main/java/com/oltpbenchmark/ThreadBench.java
```diff
@@ -420,9 +420,9 @@ private void calculateNext() {
             while (sample != null && sample.getStartNanosecond() < endNanoseconds) {
 
                 // Check if a TX Type filter is set, in the default case,
```
lmwnshn, line 422:
```suggestion
                // Check if a TxType filter is set, in the default case,
```

for consistency with below

src/main/java/com/oltpbenchmark/Phase.java
```diff
@@ -187,7 +182,7 @@ public int chooseTransaction(boolean isColdQuery) {
             for (int i = 0; i < this.weightCount; i++) {
                 weight += weights.get(i);
                 if (randomPercentage <= weight) {
-                    return i + 1;
+                    return i;
```
lmwnshn, line 185: Is this a bugfix?

lmwnshn (commented):
Personal checkpoint for the day; have skimmed everything except these five files which have more substantial changes.

```
DBWorkload
WorkloadConfiguration
BenchmarkModule
StatementDialects
TPCDSLoader
```

Overall LGTM so far, nice refactor!

src/main/java/com/oltpbenchmark/benchmarks/twitter/TwitterLoader.java
timveil: ah so here is the deal... when i reconfigured the abstract base test case to apply a smaller default scale factor to all tests (eg `.01`) i noticed some odd behavior in the twitter test.  my comment at the time said this... `fixing bug in twitter loader that only showed up with very low scale factors causing infinite loop`.  you can see that prior to this change it would have been possible for `f` to never have been incremented thus causing the infinite loop.

src/main/java/com/oltpbenchmark/Phase.java
timveil: i've broken something here that needs to be fixed before this PR can be accepted.  I need some help to better understand phases.  I think i've broken the link between `weight` and `transaction` by introducing transactionid collision.  its seems like we could simplify things here but i'll need some guidance

lmwnshn: I probably won't be able to take a more detailed pass on this until the weekend

viktor-51:
Hi I have tried your code out and I seem to have a bug using Wikipedia. When i load the data base with PostgreSQL as;

java -jar benchbase.jar -w config/workload/tpcc.xml -d config/db/postgres.xml --create=true --load=true

Then execute as;

java -jar benchbase.jar -w config/workload/tpcc.xml -d config/db/postgres.xml --execute=true

Then the database just freezes at;

[INFO ] 2022-04-23 14:38:53,269 [main]  com.oltpbenchmark.ThreadBench runRateLimitedMultiPhase - TERMINATE :: Waiting for all terminals to finish ..

But when i run the database as;

java -jar benchbase.jar -w config/workload/tpcc.xml -d config/db/postgres.xml --create=true --load=true --execute=true

Then it works fine.

Best regards Viktor Olsson :)

PR #138, opened 2022-03-18
Change so Wikipedia benchmark will work with CockroachDB's SQL statement size limit

mgoddard:
The Wikipedia benchmark involves SQL statements which are large.  Often, when batched, these will exceed the 16 MB default single statement limit set by CockroachDB (https://www.cockroachlabs.com/docs/stable/known-limitations.html#size-limits-on-statement-input-from-sql-clients).  This change allows for a limit to be placed on the length of text inserted into the OLD_TEXT column of the TEXT table, and this value is included in the XML configuration file config/cockroachdb/sample_wikipedia_config.xml: `<oldtextmaxlen>1048576</oldtextmaxlen>`.  This, coupled with a change to the default value of batch size, to 4 (`<batchsize>4</batchsize>`), should be sufficient to get the data load running reliably without exceeding the 16 MB max. statement size limit.

If this is not set here, it defaults to `Integer.MAX_VALUE`.

After making these changes, the workload appears to run at a scale factor of 500, though there may be other factors I haven't seen which factor into this.

### DIFF
--- a/config/cockroachdb/sample_wikipedia_config.xml
+++ b/config/cockroachdb/sample_wikipedia_config.xml
@@ -8,10 +8,13 @@
     <username>root</username>
     <password></password>
     <isolation>TRANSACTION_SERIALIZABLE</isolation>
-    <batchsize>128</batchsize>
+    <batchsize>4</batchsize>
 
     <!-- Scale factor is the number of wikipages *1000 -->
     <scalefactor>1</scalefactor>
+
+    <!-- Limit on the size of the string inserted into the OLD_TEXT column -->
+    <oldtextmaxlen>1048576</oldtextmaxlen>
 
     <!-- The workload -->
     <terminals>1</terminals>
--- a/config/cockroachdb/sample_wikipedia_config.xml
+++ b/config/cockroachdb/sample_wikipedia_config.xml
@@ -8,10 +8,13 @@
     <username>root</username>
     <password></password>
     <isolation>TRANSACTION_SERIALIZABLE</isolation>
-    <batchsize>128</batchsize>
+    <batchsize>8</batchsize>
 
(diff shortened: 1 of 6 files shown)


config/cockroachdb/sample_wikipedia_config.xml
```diff
@@ -8,11 +8,14 @@
...
     <password></password>
     <isolation>TRANSACTION_SERIALIZABLE</isolation>
-    <batchsize>128</batchsize>
+    <batchsize>4</batchsize>
 
     <!-- Scale factor is the number of wikipages *1000 -->
     <scalefactor>1</scalefactor>
 
+    <!-- Limit on the size of the string inserted into the OLD_TEXT column -->
+    <oldtextmaxlen>1048576</oldtextmaxlen>
```
timveil, line 17: the challenge with this approach is the xml and ultimate WorkloadConfiguration class is used by every benchmark.  This property only makes sense for wikipedia.  I'd like to find a solution thats more generic.  One thought is to handle this in the ddl for cockroach by making the column length smaller and inspecting the column length in the java code for the loader.  Another approach might be to create a facility for benchmark specific customizations.  Lets chat.

mgoddard: @timveil, after reading up a bit on this CockroachDB setting, `sql.conn.max_read_buffer_message_size`, it seems like the best approach to overcoming this 16 MB limitation per statement.  I've just dialed that up to 256 MB and am seeing fine results with a Wikipedia batch size of 32, so I think we can disregard this PR.

Thanks for having a look at it!

lmwnshn:
Closing per @mgoddard comment "I think we can disregard this PR" above

PR #144, opened 2022-03-30
Adding Missing Test cases and test case refactoring

timveil:
The test package was missing a number of benchmark tests.  This PR attempts to resolve all of that by adding any missing benchmark, loader or worker tests.  To simplify the creation of these tests, some refactoring was done to the base "abstract" test classes.  2 tests, chbenchmark and twitter, have been temporarily `@Ignore`d as the changes required to make them work have already been addressed in PR #134 which. 

Of course when adding the missing test cases, some issues were identified in either the benchmarks themselves (usually having to do with behavior at very low SF, or with the HSQLDB dialect files).  These issues were resolved as part of this PR. 

Furthermore, I changed the way HSQLDB is initialized so that tests can be paused and the database can be accessed/queried.  This is extremely useful for troubleshooting.


timveil:
i need to do one more pass over these changes at a very low scale factor `.001` as now the full suite of unit tests at SF `.01` takes nearly 20 minutes (most being TPCH).  i'll ping ya'll for a review once i've finished any changes.

lmwnshn:
Please ping me on Slack (otherwise I might miss it). Thanks!

src/main/java/com/oltpbenchmark/benchmarks/noop/procedures/NoOp.java
```diff
@@ -38,7 +38,7 @@
 
     // The query only contains a semi-colon
     // That is enough for the DBMS to have to parse it and do something
-    public final SQLStmt noopStmt = new SQLStmt(";");
+    public final SQLStmt noopStmt = new SQLStmt("select 1");
```
apavlo, line 41: This is not the same as a NOOP since it requires evaluation. Is there a reason for this change?

Maybe we should rename it to the "DoNothing" benchmark and then have separate `SelectOne` and `SemiColon` procedures.

apavlo (changes requested):
I disagree with your change to NoOp.

src/main/java/com/oltpbenchmark/benchmarks/noop/procedures/NoOp.java
timveil: NoOp was one of the missing test cases; unfortunately HSQLDB doesn't accept `;` as a valid sql statement.  my thinking was this was a reasonable fix but now that i'm looking at it i should have just done this with a hsqldb dialect file and not changed the java class.  is that a better approach?

src/test/java/com/oltpbenchmark/benchmarks/chbenchmark/TestCHBenCHmarkWorker.java
```diff
@@ -0,0 +1,38 @@
...
+
+package com.oltpbenchmark.benchmarks.chbenchmark;
+
+import com.oltpbenchmark.api.AbstractTestWorker;
+import com.oltpbenchmark.api.Procedure;
+import org.junit.Ignore;
+
+import java.util.List;
+
+@Ignore("the testcase is under development")
```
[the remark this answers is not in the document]
timveil: see above.  i need to ignore this until config-refactor is merged, then i can re-enable.

src/test/java/com/oltpbenchmark/benchmarks/twitter/TestTwitterWorker.java
```diff
@@ -0,0 +1,21 @@
+package com.oltpbenchmark.benchmarks.twitter;
+
+import com.oltpbenchmark.api.AbstractTestWorker;
+import com.oltpbenchmark.api.Procedure;
+import org.junit.Ignore;
+
+import java.util.List;
+
+@Ignore("the testcase is under development")
```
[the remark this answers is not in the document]
timveil: see above

src/main/java/com/oltpbenchmark/benchmarks/noop/procedures/NoOp.java
timveil: this has been resolved by reverting the change to NoOp.java and instead adding a dialect-hsqldb.xml file with a hsqldb friendly query

apavlo (approved):
LGTM. Please resolve conflicts

PR #147, opened 2022-04-07
Add TPC-H support for SQL Server

jcamachor:


src/main/java/com/oltpbenchmark/catalog/Table.java
```diff
@@ -91,7 +91,7 @@ public void addIndex(Index index) {
 
     public Index getIndex(String indexName) {
         for (Index catalog_idx : this.indexes) {
-            if (catalog_idx.getName().equalsIgnoreCase(indexName)) {
+            if (indexName.equalsIgnoreCase(catalog_idx.getName())) {
```
apavlo, line 94: Why does this matter?

src/main/resources/benchmarks/tpch/dialect-sqlserver.xml
```diff
@@ -0,0 +1,547 @@
+<?xml version="1.0"?>
+<dialects>
+    <dialect type="SQLSERVER">
+        <procedure name="Q1">
+            <statement name="query_stmt">
+                select
```
apavlo, line 6: Can you pretty-print these queries?

config/sqlserver/sample_tpch_config.xml
```diff
@@ -0,0 +1,143 @@
...
+    <password>P@ssw0rd</password>
+    <isolation>TRANSACTION_SERIALIZABLE</isolation>
+    <batchsize>128</batchsize>
+
+    <!-- Location for the files generated by tpch dbgen -->
+    <datadir>data/tpch-sf0.01</datadir>
+
+    <!-- Format of the files that contain the tpch data -->
+    <!-- Values: csv or tbl-->
+    <fileFormat>tbl</fileFormat>
```
apavlo, line 18: You can remove these parameters. They're no longer needed.

apavlo (changes requested):
Minor quibbles.

apavlo:
@jcamachor Thanks for sending this. We should probably be testing MSSQL in our Github actions. Is this something you could help with?

src/main/java/com/oltpbenchmark/catalog/Table.java
jcamachor: When index type is `tableIndexStatistic`, index name is null, and we end up hitting a NPE.

I just worked around the issue at the time without exploring it much and I didn't even realize that this change was part of the PR :) I was examining the [DatabaseMetaData interface](https://docs.oracle.com/javase/7/docs/api/java/sql/DatabaseMetaData.html#getIndexInfo(java.lang.String,%20java.lang.String,%20java.lang.String,%20boolean,%20boolean)) definition and it seems this is the expected behavior by all implementers:
```
INDEX_NAME String => index name; null when TYPE is tableIndexStatistic
```
I think the proper solution should be ignoring the entry in the loop if type is `tableIndexStatistic`. I'll change the code. Is it fine to do it as part of this PR or you prefer I open a new one?

src/main/resources/benchmarks/tpch/dialect-sqlserver.xml
jcamachor: Done. Please let me know if I followed the right convention (looked at a few other files).

jcamachor:
@apavlo , sure, I can do that in a follow-up PR, I can use one of the images in https://hub.docker.com/_/microsoft-mssql-server (already doing some quick tests using one of those) and enable a few different benchmarks.

src/main/java/com/oltpbenchmark/catalog/Table.java
apavlo: You can do it in this one.

bpkroth:
> @jcamachor Thanks for sending this. We should probably be testing MSSQL in our Github actions. Is this something you could help with?

I added that in this PR to @jcamachor's branch: <https://github.com/jcamachor/benchbase/pull/1>
Also added TPC-C support to CI checks.  Can probably add others at a later date as well if desired.

apavlo:
@jcamachor @bpkroth Everything looks good except for the symlink to the DDL file. I think that it is better to have a single file for now and then comment out the column store stuff. Then create issue about being able to specify/override the DDL file from the config file.

apavlo (approved):
LGTM

PR #150, opened 2022-04-13
QA-3470 TPC-C/TPC-H on Ignite2

Berkof:


apavlo:
@Berkof Why close this??? If it's legit we'll take it.

PR #152, opened 2022-04-19
Fix infinite loop on retry logic on transaction failure #152

anjagruenheid:
Fixes bug that causes an endless loop when running into an error during query execution.


apavlo:
@anjagruenheid Thanks for sending this. I think the logic is correct. What DBMS / benchmark did you hit this issue?

anjagruenheid:
We ran into it when trying to run TPC-H on SQL Server before we had the custom dialect that Jesus' PR adds.

PR #163, opened 2022-05-13
Run benchbase from any directory

ajmaidak:
See: https://github.com/cmu-db/benchbase/pull/111

This brings back this change with a fallback to check for the
config/plugin.xml so that benchbase can still be executed from the
project root using the mvn wrapper.

### DIFF
--- a/src/main/java/com/oltpbenchmark/DBWorkload.java
+++ b/src/main/java/com/oltpbenchmark/DBWorkload.java
@@ -61,7 +61,15 @@ public class DBWorkload {
         // create the command line parser
         CommandLineParser parser = new DefaultParser();
 
-        XMLConfiguration pluginConfig = buildConfiguration("config/plugin.xml");
+        // load the plugin configuration from the jar's folder or if not found the the $cwd/config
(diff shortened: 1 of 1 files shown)


src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -61,7 +61,15 @@ public static void main(String[] args) throws Exception {
...
-        XMLConfiguration pluginConfig = buildConfiguration("config/plugin.xml");
+        // load the plugin configuration from the jar's folder or if not found the the $cwd/config
+        String jarDirPath = new File(DBWorkload.class.getProtectionDomain().getCodeSource().getLocation().toURI()).getParent();
+        File pluginFile = new File(jarDirPath + "/config/plugin.xml");
+        XMLConfiguration pluginConfig;
+        if (pluginFile.exists()) {
+            pluginConfig = buildConfiguration(jarDirPath + "/config/plugin.xml");
+        } else {
+            pluginConfig = buildConfiguration("config/plugin.xml");
+        }
```
bpkroth, line 72: I suggest reversing this logic.
Current state of affairs is to load the data from the current directory, so we should default to that to avoid breaking existing installs that might be doing something clever.
If that *doesn't* exist, then we should fallback to loading it from the jar directory.

PR #164, opened 2022-05-20
Fix serial execution start and reset.

anjagruenheid:
Noticed wrong execution of serial TPC-H benchmark due to a) state being set to WARMUP even though serial execution does not require a warm-up phase, b) workers being started right after initialization due to single-threaded execution, and c) an incorrect reset for serial after a phase has finished. This resulted in unnecessary query executions during the warm-up phase and skipped cold queries for Q1 in all phases after the first one.
-- Added a synchronized state adjustment for serial execution to set state to COLD_QUERY before workers are started.
-- Added a synchronized state adjustment when execution is reset between phases.

### DIFF
--- a/src/main/java/com/oltpbenchmark/ThreadBench.java
+++ b/src/main/java/com/oltpbenchmark/ThreadBench.java
@@ -15,7 +15,6 @@
  *
  */
 
-
 package com.oltpbenchmark;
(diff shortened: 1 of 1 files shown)


apavlo (approved):
LGTM

PR #166, opened 2022-05-23
Tpch load retries

dbanda:
In our testing with TPCH we found that for large scale factors the loader wasn't fault tolerant and would have to be restarted after any connectivity failure. In this PR we implement a retry mechanism where the loader retries batches a few times after failure.  This was very helpful in cases where we had very high scale factor and there was intermittent connectivity. It's quite a big PR so feedback on how I can package it better is greatly appreciated.

### DIFF
--- a/src/main/java/com/oltpbenchmark/benchmarks/tpch/TPCHLoader.java
+++ b/src/main/java/com/oltpbenchmark/benchmarks/tpch/TPCHLoader.java
@@ -42,15 +42,14 @@ import com.oltpbenchmark.catalog.Table;
 
 import java.sql.*;
 import java.util.ArrayList;
+import java.util.LinkedList;
 import java.util.List;
 import java.util.concurrent.CountDownLatch;
 
 public class TPCHLoader extends Loader<TPCHBenchmark> {
-    private WorkloadConfiguration configuration;
 
     public TPCHLoader(TPCHBenchmark benchmark) {
         super(benchmark);
-        this.configuration = benchmark.getWorkloadConfiguration();
     }
 
     private enum CastTypes {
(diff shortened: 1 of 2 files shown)


apavlo:
Thanks for sending this. We will take a look at it later this week.

src/main/java/com/oltpbenchmark/benchmarks/tpch/TPCHLoader.java
```diff
@@ -153,45 +156,55 @@ public List<LoaderThread> createLoaderThreads() {
...
 
             @Override
             public void afterLoad() {
                 regionLatch.countDown();
             }
+
+            @Override
+            public PreparedStatement getPreparedStatement(Connection conn) throws SQLException{
+                return getInsertStatement(conn, TABLENAME_REGION);
+            }
```
apavlo, line 175: Again, I don't understand the purpose of this change?

src/main/java/com/oltpbenchmark/benchmarks/tpch/TPCHLoader.java
```diff
@@ -335,49 +373,90 @@ public void beforeLoad() {
...
+                                        prepStmt.close();
+                                        conn = this.benchmark.makeConnection();
+                                        prepStmt = loader.getPreparedStatement(conn);
+                                        for (List<Object> retryElems : currentBatch){
+                                            addElems(prepStmt, retryElems, types);
+                                            prepStmt.addBatch();
+                                        }
+                                    }
+                                    LOG.debug("retrying {} times batch {} for table {}", retryCount, recordsRead, tableName);
+                                    Thread.sleep((long) (500+500*Math.random()));
```
apavlo, line 451: Why are you sleeping?

src/main/java/com/oltpbenchmark/benchmarks/tpch/TPCHLoader.java
```diff
@@ -335,49 +373,90 @@ public void beforeLoad() {
...
+                            }catch (Exception e) {
+                                LOG.error(e.getMessage(), e);
+                                
+                                retryCount++;
+                                if (retryCount == maxRetryCount){
+                                    currentBatch.clear();
+                                    LOG.error("Failed batch {} for table {}", recordsRead, tableName);
+                                }
+                                else{
+                                    if (!prepStmt.getConnection().isValid(1000)) {
```
apavlo, line 441: What is this?

src/main/java/com/oltpbenchmark/benchmarks/tpch/TPCHLoader.java
```diff
@@ -335,49 +373,90 @@ public void beforeLoad() {
...
+                                        prepStmt.close();
+                                        conn = this.benchmark.makeConnection();
+                                        prepStmt = loader.getPreparedStatement(conn);
+                                        for (List<Object> retryElems : currentBatch){
+                                            addElems(prepStmt, retryElems, types);
+                                            prepStmt.addBatch();
+                                        }
+                                    }
```
apavlo, line 449: What is this logic? Why are you making a new connection? Why not have the retry logic outside of this function so that you can catch, rebuild the batch, and then try again?

apavlo (changes requested):
Please provide additional comments to explain your new logic.
Retrying the batch from inside the exception handler seems messy.

PR #173, opened 2022-05-31
Fix for WikipediaLoader on small scalefactor

### LINKED ISSUE #171: Wikipedia Title Generation is Broken
There is a disconnect between how we generate page titles during loading and execution using `WikipediaUtil.generatePageTitle()`. For a given `page_id`, we cannot guarantee that it will always generate the same title. Thus, all the invocations of `GetPage` will fail because the page title will never match.

We need to make this method deterministic for a given `page_id`.

https://github.com/cmu-db/benchbase/blob/main/src/main/java/com/oltpbenchmark/benchmarks/wikipedia/util/WikipediaUtil.java#L28

apavlo:
Fixes #171

1. We weren't loading the last batch for the REVISIONS table.
2. The page title generation code was broken so all the GetPage txns were failing. 

https://youtu.be/BB0gUb5Yo9A


bpkroth:
@apavlo was there more you wanted to fix here?
It seems to be working well enough to me.
There are still some duplicate key warnings returned, but that seems to be somewhat expected perhaps?

Anyways, I have another PR in #268 that builds on this to also check that we aren't letting a bunch of errors through on the benchmarks in the future.

We can either merge this one first and then that one, or the other one and mark this one as superseded by it.

Let me know what you think.  Thanks!

bpkroth:
> @apavlo was there more you wanted to fix here? It seems to be working well enough to me. There are still some duplicate key warnings returned, but that seems to be somewhat expected perhaps?
> 
> Anyways, I have another PR in #268 that builds on this to also check that we aren't letting a bunch of errors through on the benchmarks in the future.
> 
> We can either merge this one first and then that one, or the other one and mark this one as superseded by it.
> 
> Let me know what you think. Thanks!

Nm, I see the problem now.

The unit tests fail because HSQL doesn't have an auto increment column for old_id in UpdatePage.  Will need to take another look at that next week ...

bpkroth:
> > @apavlo was there more you wanted to fix here? It seems to be working well enough to me. There are still some duplicate key warnings returned, but that seems to be somewhat expected perhaps?
> > Anyways, I have another PR in #268 that builds on this to also check that we aren't letting a bunch of errors through on the benchmarks in the future.
> > We can either merge this one first and then that one, or the other one and mark this one as superseded by it.
> > Let me know what you think. Thanks!
> 
> Nm, I see the problem now.
> 
> The unit tests fail because HSQL doesn't have an auto increment column for old_id in UpdatePage. Will need to take another look at that next week ...

Fixing this mostly by adding custom ddl for hsqldb.  Changes are in #268

bpkroth:
Included this and additional unit test fixups in #268

PR #204, opened 2022-07-06
Fix bugs with TPCH implementation

TheoVanderkooy:
This changes fixes a few bugs with the TPCH queries where it does not match the [specification](https://www.tpc.org/tpc_documents_current_versions/pdf/tpc-h_v2.17.1.pdf). Note: each fix is a separate commit so they can be reviewed independently if that is easier. (I can also split into multiple PRs if preferred, but the changes are pretty small)

1. For Q16, Q17, and Q19: The brand name in the query is capitalised wrong. It should be 'Brand#MN' (and this is what is generated when loading the data) but was using all-caps 'BRAND#MN'.
2. For Q12: The query text is wrong in the java code. (it had Q11's query text) The XML files for different "dialects" had the correct query which is why this didn't error due to mismatched number & types of query parameters. (Maybe there is some database where this does produce an error, but I have only been using it with postgres)
3. For Q11 there are 2 fixes. First it had 'ETHIOPIA' hardcoded for one of the parameters which should be chosen randomly and match the other [NATION] parameter. Second, this query is supposed to depend on the scale factor but it was previously ignoring this (assuming SF=1). To make the scale factor available I stored the workload configuration in the `Procedure` base class during `initialize` instead of just the database type, but if there is a better way to access the scale factor I am happy to change this.

### DIFF
--- a/src/main/java/com/oltpbenchmark/api/Procedure.java
+++ b/src/main/java/com/oltpbenchmark/api/Procedure.java
@@ -17,7 +17,6 @@
 
 package com.oltpbenchmark.api;
 
-import com.oltpbenchmark.WorkloadConfiguration;
 import com.oltpbenchmark.jdbc.AutoIncrementPreparedStatement;
 import com.oltpbenchmark.types.DatabaseType;
 import org.slf4j.Logger;
@@ -37,7 +36,7 @@ public abstract class Procedure {
     private static final Logger LOG = LoggerFactory.getLogger(Procedure.class);
 
     private final String procName;
-    protected WorkloadConfiguration workConf;
+    private DatabaseType dbType;
     private Map<String, SQLStmt> name_stmt_xref;
 
     /**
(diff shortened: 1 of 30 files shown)


src/main/java/com/oltpbenchmark/api/Procedure.java
```diff
@@ -51,11 +52,12 @@ protected Procedure() {
...
      *
      * @param <T>
+     * @param workConf Workload configuration
      * @return
      */
     @SuppressWarnings("unchecked")
-    protected final <T extends Procedure> T initialize(DatabaseType dbType) {
-        this.dbType = dbType;
+    protected final <T extends Procedure> T initialize(WorkloadConfiguration workConf) {
+        this.workConf = workConf;
```
apavlo, line 60: I don't think we should do this. It makes it more brittle since `WorkloadConfiguration` is basically a dumping ground for everything. Let's revert this and only pass in `DatabaseType`.

src/main/java/com/oltpbenchmark/benchmarks/tpch/procedures/Q11.java
```diff
@@ -63,12 +63,12 @@ protected PreparedStatement getStatement(Connection conn, RandomGenerator rand)
         String nation = TPCHUtil.choice(TPCHConstants.N_NAME, rand);
 
         // FRACTION is chosen as 0.0001 / SF
-        // TODO: we should technically pass dbgen's SF down here somehow
-        double fraction = 0.0001;
+        double fraction = 0.0001 / this.workConf.getScaleFactor();
```
apavlo, line 66: We should pass this in as an input parameter to `getStatement()`

apavlo (changes requested):
This is great! Can you make it so that we pass in the scale factor as an input parameter to the `getStatement()` methods instead of storing the `WorkloadConfiguration` in `Procedure`?

src/main/java/com/oltpbenchmark/api/Procedure.java
TheoVanderkooy: Done.

apavlo (approved):
LGTM. I will merge when it passes.

PR #208, opened 2022-07-12
Add `newConnectionPerTxn` config option and implementation

mbutrovich:
[pgbench](https://www.postgresql.org/docs/current/pgbench.html) has `-c` or `--connect` flag with the following behavior:
> Establish a new connection for each transaction, rather than doing it just once per client session. This is useful to measure the connection overhead.

This PR attempts to replicate that behavior with a `newConnectionPerTxn` config option.

This might be my first BenchBase PR so please double-check me on naming/code conventions for this repo. I'll also look into the test infrastructure a bit to see how easy it is to add a test scenario for this, so consider this PR WIP for the moment.

### DIFF
--- a/src/main/java/com/oltpbenchmark/api/Worker.java
+++ b/src/main/java/com/oltpbenchmark/api/Worker.java
@@ -50,7 +50,7 @@ public abstract class Worker<T extends BenchmarkModule> implements Runnable {
 
     private final int id;
     private final T benchmark;
-    protected final Connection conn;
+    protected Connection conn = null;
     protected final WorkloadConfiguration configuration;
     protected final TransactionTypes transactionTypes;
     protected final Map<TransactionType, Procedure> procedures = new HashMap<>();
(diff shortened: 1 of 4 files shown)


src/main/java/com/oltpbenchmark/api/Worker.java
```diff
@@ -74,12 +74,14 @@ public Worker(T benchmark, int id) {
...
 
-        try {
-            this.conn = this.benchmark.makeConnection();
-            this.conn.setAutoCommit(false);
-            this.conn.setTransactionIsolation(this.configuration.getIsolationMode());
-        } catch (SQLException ex) {
-            throw new RuntimeException("Failed to connect to database", ex);
+        if (!this.configuration.getNewConnectionPerTxn()) {
+            try {
+                this.conn = this.benchmark.makeConnection();
```
apavlo, line 79: Can you refactor `BenchmarkModule.makeConnection()` so that it takes in the autoCommit + txnIsolation level as arguments? That way we don't have dupe code for your new feature.

src/main/java/com/oltpbenchmark/api/Worker.java
```diff
@@ -430,6 +446,14 @@ protected final void doWork(DatabaseType databaseType, TransactionType transacti
                     }
 
                 } finally {
+                    if (this.configuration.getNewConnectionPerTxn() && this.conn != null) {
+                        try {
+                            this.conn.close();
+                            this.conn = null;
+                        } catch (SQLException e) {
+                            LOG.error("Connection couldn't be closed.", e);
```
apavlo, line 454: When would this happen? This seems like it should be a fatal error?

apavlo (changes requested):
Minor requests

src/main/java/com/oltpbenchmark/api/Worker.java
mbutrovich: I just duplicated the behavior in `tearDown()`. Presumably the `conn.close()` call can throw which is why the logic is the way it is in `tearDown()`.

src/main/java/com/oltpbenchmark/api/Worker.java
apavlo: Ignore this statement. I see why you didn't do this.

apavlo (approved):
LGTM

PR #211, opened 2022-07-20
null check `postPhase` in `Worker.run` to resolve race condition at terminate

### LINKED ISSUE #37: Find and fix race condition during post-phase checks.
#36 patches over a race condition where the state disappears and you get a null pointer exception.

I haven't looked into possible causes yet.

mbutrovich:
`ThreadBench` and `WorkloadState` rely on `WorkloadState.currentPhase` being `null` as the indicator that the benchmark is over. For example: https://github.com/cmu-db/benchbase/blob/1c43cad2a9b47611806f1e101d7e4b636ae8f105/src/main/java/com/oltpbenchmark/WorkloadState.java#L205-L206

However, `Worker` relies on `BenchmarkState.state` being `DONE` as its indicator that the benchmark is over. For example, this switch assumes that because state is still `MEASURE` that the benchmark is still running and that the result of `getCurrentPhase` is safe to access: https://github.com/cmu-db/benchbase/blob/1c43cad2a9b47611806f1e101d7e4b636ae8f105/src/main/java/com/oltpbenchmark/api/Worker.java#L288-L296

The race occurs between the previously mentioned switch in `Worker` and in `ThreadBench` where this line advances to the next phase, which in the case of the benchmark ending, puts the current phase to `null`. If a `Worker` tries to read the current phase before the state is set to `DONE` when `ThreadBench` calls `startCoolDown()` then benchbase goes boom. https://github.com/cmu-db/benchbase/blob/1c43cad2a9b47611806f1e101d7e4b636ae8f105/src/main/java/com/oltpbenchmark/ThreadBench.java#L222-L229

I believe the easiest fix here is a documented `null` check in `Worker`. Maybe there's a better way to sync up the indicators of the benchmark ending between all of these classes, but this seems like the lowest risk change.

Fixes #37.

### DIFF
--- a/src/main/java/com/oltpbenchmark/api/Worker.java
+++ b/src/main/java/com/oltpbenchmark/api/Worker.java
@@ -378,7 +378,7 @@ public abstract class Worker<T extends BenchmarkModule> implements Runnable {
      * implementing worker should return the TransactionType handle that was
      * executed.
      *
-     * @param databaseType    TODO
+     * @param databaseType TODO
      * @param transactionType TODO
      */
     protected final void doWork(DatabaseType databaseType, TransactionType transactionType) {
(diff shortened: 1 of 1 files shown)


apavlo (approved):
LGTM

PR #214, opened 2022-07-29
Add Sleep benchmark and dialects for MySQL and Postgres

mbutrovich:
WIP PR. Adds a new benchmark (dialects for Postgres and MySQL so far) to just sleep.
Questtions:
- Not sure how to write the generic query for this since it's not in the SQL standard. Right now it's just a `;` like in NoOp.
- Do we need a DDL file? I borrowed the one from NoOp.


apavlo:
@mbutrovich Are you going to add XML config options to control # of queries per txn and amount of sleep per query?

PR #215, opened 2022-07-29
Container Publishing and Build Paralellism

bpkroth:
This change set handles publishing docker containers to an upstream docker registry (currently ACR) after successful builds for approved PRs merged to `main`.  To account for security fixes in the container images, a nightly build job was also added.

The PR also splits the published containers into `benchbase-dev` with a JDK for the devcontainer and `benchbase` with a JRE for the runtime only version (previously both contained a full JDK).  The later contains all profiles.  Alternatively, there are singleton profile images (e.g. `benchbase-postgres`).  This, along with some internal file deduplication, improves container size.
Additionally, some tweaks were made to allow building each profile in parallel within the devcontainer.

While doing that a race in test target execution was noticed when multiple maven threads are allowed to run.
For now, we have simply disabled parallelism for the test target.

Currently, the images are published as `latest` however, can also be generated for tagged `v2022` style releases.

Some documentation and CI tests are also included.


pom.xml
```diff
@@ -15,6 +15,7 @@
         <java.version>17</java.version>
         <maven.compiler.source>17</maven.compiler.source>
         <maven.compiler.target>17</maven.compiler.target>
+        <buildDirectory>${project.basedir}/target</buildDirectory>
```
bpkroth, line 18: This is the default build output directory and is referenced later.
However, providing it as a property allows us to override it at the command line to allow for parallel builds.

apavlo (approved):
LGTM

docker/benchbase/devcontainer/build-in-container.sh
```diff
@@ -0,0 +1,107 @@
...
+
+# Make sure that we've built the base stuff first before we start.
+mvn -T 2C -B --file pom.xml process-resources compile # ${TEST_TARGET:-}
+if [ -n "${TEST_TARGET:-}" ]; then
+    # FIXME: Run tests without parallelism to work around some buggy behavior.
+    mvn -B --file pom.xml $TEST_TARGET
+fi
+
+for profile in ${BENCHBASE_PROFILES}; do
+    build_profile "$profile" &
```
bpkroth, line 92: I think there's a race in maven downloading required artifacts here when doing this in parallel. Individual instances are downloading files as .part and then moving them into place, but when multiple instances download the same file one instance can move the file out of place before the other and cause a file not found error.
Think I have a fix. Will test shortly.

bpkroth: Did this. Basically just run a maven target that tries to download most of the resources for each profile serially first.
Should probably also report this as a bug to maven upstream...

bpkroth:
I will ping you out of band to setup the secrets for ACR login access and then I think this can get squash merged.

.github/workflows/maven.yml
```diff
@@ -381,25 +400,48 @@ jobs:
...
-          cat "config/postgres/sample_${benchmark}_config.xml" | sed -e 's/localhost:5432/postgres:5432/g' > /tmp/config.xml
-          # Lookup the service container's docker network so we can place the benchbase container in it too.
-          docker_network="$(docker ps --filter expose=5432 --format '{{.Networks}}')"
-          # Map the adjusted config into the container and use it to run the test.
-          mkdir -p results
-          docker run --rm --name benchbase-postgres --network "$docker_network" \
-            --env BENCHBASE_PROFILE=postgres -v /tmp/config.xml:/tmp/config.xml -v "$PWD/results:/benchbase/results" \
-            benchbase:latest -b "$benchmark" -c /tmp/config.xml --create=true --load=true --execute=true
-          ls results/${benchmark}_*.csv
+          for image in benchbase benchbase-postgres; do
```
bpkroth, line 415: Just a note for later:
One could refactor this a tad to build/test the singleton images as a parallel matrix strategy is done for the original build/testing workflows.
We already do that in a way that works so I opted for a smaller set of tests here for now.
One caveat for that if we wanted to change it in the future would be to separate out the shared `benchbase` image that builds all profiles into it so that the other jobs only push their own `benchbase-{profile}` image (as it currently stands those jobs would simultaneously produce a `benchbase` image that only has the single profile in it.

bpkroth: I added a tweak for that too actually: 72aa4126

PR #217, opened 2022-08-12
Add Egress benchmark

mbutrovich:
Not ready for review yet.

This adds a configurable Egress benchmark to return result rows based on two parameters: `egress_tuple_bytes` and `egress_num_tuples`. Since UDF syntax varies by DBMS, this only includes a dialect and ddl file for Postgres. Hopefully folks can add more if this merges.

### DIFF
--- a/.github/workflows/maven.yml
+++ b/.github/workflows/maven.yml
@@ -222,7 +222,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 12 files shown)


mbutrovich:
Apparently the `ScriptRunner` class we use to run ddl .sql files can't handle plpgsql. I have to imagine there's a better way to load and run a .sql file over JDBC than a Java class from 2004, but I'll hack around it locally for now by just issuing a `psql` command with the .sql file.

PR #221, opened 2022-09-01
changes made to featurebench

GargiChakraverty-yb:


apavlo:
@GargiChakraverty-yb If this is a new benchmark workload, we would definitely be interested in it.

PR #222, opened 2022-09-06
Fix MySQL ER_LOCK_WAIT_TIMEOUT SQLState code.

### LINKED ISSUE #220: Incorrect SQLState value in Worker.java
In Worker when a retryable exception occurs the sqlstate and the error code are checked to see if the transaction is retryable or not. 
The sql state and error code for the lock wait timeout exception are written as 41000 and 1205 respectively.
According to the mysql link https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-error-sqlstates.html, the correct sqlstate is 40001.

AroraAsh:
Fix #220.


lmwnshn (approved):
LGTM, thanks for the PR!

https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-reference-error-sqlstates.html

PR #228, opened 2022-10-24
Remove 'data' directory

apavlo:
I think we can safely remove these files.


bpkroth:
The `templated` benchmark has an example config in the `data` directory right btw.

PR #234, opened 2022-11-12
Write result files after error

### LINKED ISSUE #232: Results files aren't written during error
If I run a benchmark and an error occurs in the middle no results files are written.  

For example, say this is expected to run for 1 hour with `<time>3600</time>`, and at `00:00:00` I start:
```
java -jar benchbase.jar -b tpcc -c ....
```
Say at time `00:00:59`  I get a exception:
```
Caused by: com.microsoft.sqlserver.jdbc.SQLServerException: The connection is closed.
```

In this case, nothing at all is written to file, and I lose 59 minutes of throughput/latency/etc data for that run.  This could be useful for debugging for running experiments.

Would it be possible to change things so at least partial results files are written/flushed at the end before exiting?  Maybe some flushes or intermediate writes?  Or is the concern that this would hurt performance?

jcamachor:
Fixes #232

### DIFF
--- a/src/main/java/com/oltpbenchmark/BenchmarkState.java
+++ b/src/main/java/com/oltpbenchmark/BenchmarkState.java
@@ -92,6 +92,12 @@ public final class BenchmarkState {
         state = State.MEASURE;
     }
 
+    public void signalError() {
+        // A thread died, decrement the count and set error state
+        notDoneCount.decrementAndGet();
+        state = State.ERROR;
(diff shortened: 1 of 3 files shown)


jcamachor:
To fix the issue, I tried an approach that would not require rewriting the current workflow. Please let me know what you think.
I tested locally with SQLServer and results before failure is found are reported now.

PR #242, opened 2023-01-19
Add additional benchmarks for SqlServer

bpkroth:


bpkroth:
Handling this in #264

Issue #122, opened 2022-03-03
aborted transactions in Wikipedia


timveil:
If you look at any recent run of `wikipedia` you will see that 3 of the procedures fail 100% of the time with "aborted" transactions.  Upon further inspection i think this was unintended but i'd like to confirm.

```
Aborted Transactions:
com.oltpbenchmark.benchmarks.wikipedia.procedures.UpdatePage/03                  [ 4170] ******
com.oltpbenchmark.benchmarks.wikipedia.procedures.GetPageAnonymous/04            [54069] ********************************************************************************
com.oltpbenchmark.benchmarks.wikipedia.procedures.GetPageAuthenticated/05        [  585] 
```

The cause of this appears to be that the runner logic essentially creates new page variables when executing the above procedures instead of referencing existing pages previously created.  In other words, instead of doing lookups using existing data, the code generates keys that don't already exist in the database... thus always returning 0 records.  The fix here is to ensure that the above procedures get a handle on an existing page. 

I have a fix completed and tested but want to open this issue for history and discussion if i'm missing something.

timveil:
a good example of this is here... https://github.com/cmu-db/benchbase/actions/runs/1918000781.  if you look at the `postgres` run you see the following...

```
Completed Transactions:
com.oltpbenchmark.benchmarks.wikipedia.procedures.AddWatchList/01                    [ 633] 
com.oltpbenchmark.benchmarks.wikipedia.procedures.RemoveWatchList/02             [ 609] 
com.oltpbenchmark.benchmarks.wikipedia.procedures.UpdatePage/03                          [   0] 
com.oltpbenchmark.benchmarks.wikipedia.procedures.GetPageAnonymous/04            [   0] 
com.oltpbenchmark.benchmarks.wikipedia.procedures.GetPageAuthenticated/05        [   0] 

Aborted Transactions:
com.oltpbenchmark.benchmarks.wikipedia.procedures.UpdatePage/03                         [ 4113] 
com.oltpbenchmark.benchmarks.wikipedia.procedures.GetPageAnonymous/04            [54020] 
com.oltpbenchmark.benchmarks.wikipedia.procedures.GetPageAuthenticated/05        [  626]
```

bpkroth:
Noticed this recently too.  Any updates on this?
Was thinking of adding a check to the CI pipelines to parse the results of the benchmarks and make sure things like this don't happen in the future.

PR #268, opened 2023-03-23
CI checks on benchmark histograms and associated benchmark fixups

bpkroth:
As previously noted in #122 the Wikipedia benchmark has a bug that results in most transactions getting aborted.

The fix developed in #173 by @apavlo seems to mostly fix it (though there are still some duplicate key errors generated by the benchmark's AddWatchPage procedure).

This PR pulls in those changes plus some additional ones (below) and additionally adds some CI checks on the json histogram outputs so that any benchmark the generates errors more than 1% of the completed transactions should be considered a failure (previously a CI failure was reported only if java exited non-zero, which doesn't happen when the many transactions fail, only when an unexpected exception occurs).

The 1% value is just an initial threshold and can be adjusted both on a per database and a per benchmark basis.
We do this in several places (e.g. TATP benchmark) in order to keep the changes required for this PR somewhat smaller and leave it for future work to improve those benchmarks.

To address the Wikipedia failures in some databases, we expand on auto-increment column support to SqlServer, add an auto-increment DDL schema for hsqldb for unit test purposes.

For others, we introduce improved schemas and/or dialects to fix query errors.

Note: in the case of sqlite, which currently lacks support for things like native `SLEEP` or `TABLOCK` we leave a more simplified version of the query for now and mark it for future work to improve.


docker/mysql-latest/docker-compose.yml
```diff
@@ -8,8 +8,6 @@ services:
     image: mysql:latest
     command: --max-connections=500
     environment:
-      MYSQL_USER: admin
-      MYSQL_PASSWORD: password
```
bpkroth, line 12: These are duplicates from values below.

src/main/java/com/oltpbenchmark/api/Worker.java
bpkroth, line 1: Am guessing that prettier plugin did all of these changes?

pom.xml
```diff
@@ -216,7 +216,7 @@
         <dependency>
             <groupId>org.hsqldb</groupId>
             <artifactId>hsqldb</artifactId>
-            <version>2.6.1</version>
+            <version>2.7.1</version>
```
bpkroth, line 219: Bumped the hsqldb plugin while working on debugging invalid cast exceptions.  Not required, but gonna leave it.

src/main/java/com/oltpbenchmark/benchmarks/wikipedia/procedures/UpdatePage.java
```diff
@@ -81,12 +80,12 @@ public void run(Connection conn, long textId, int pageId, String pageTitle, Stri
             int param = 1;
             ps.setInt(param++, pageId);       // rev_page
             ps.setLong(param++, nextTextId);   // rev_text_id
-            ps.setString(param++, revComment);// rev_comment
+            ps.setString(param++, revComment.substring(0, Math.min(revComment.length(), 255-1))); // rev_comment
```
bpkroth, line 83: Avoids some "Unexpected Sql Errors" due to possible content truncation.  Since this is all auto generated content anyways, I think it's fine to just truncate it in the client and avoid the server error.

src/main/java/com/oltpbenchmark/types/DatabaseType.java
```diff
@@ -45,9 +45,9 @@ public enum DatabaseType {
     ORACLE(true, false),
     SINGLESTORE(true, false),
     SPANNER(false, true),
-    SQLAZURE(true, false),
+    SQLAZURE(true, true),
     SQLITE(true, false),
-    SQLSERVER(true, false),
+    SQLSERVER(true, true),
```
bpkroth, line 50: Need to include the column names in the `INSERT` statement for the `SET INSERT_IDENTITY [table_name] ON` command to work for the session.

src/main/resources/benchmarks/wikipedia/dialect-sqlserver.xml
```diff
@@ -29,6 +29,12 @@
             </statement>
         </procedure>
         <procedure name="UpdatePage">
+            <statement name="insertRevision">
+                INSERT INTO revision (rev_page, rev_text_id, rev_comment, rev_minor_edit, rev_user, rev_user_text, rev_timestamp, rev_deleted, rev_len, rev_parent_id) OUTPUT INSERTED.rev_id VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+            </statement>
+            <statement name="insertText">
+                INSERT INTO text (old_page,old_text,old_flags) OUTPUT INSERTED.old_id VALUES (?,?,?)
```
bpkroth, line 36: Need to use `OUTPUT` clause to capture the generated column id.

.github/workflows/maven.yml
```diff
@@ -85,6 +86,7 @@ jobs:
     needs: package-and-upload
     runs-on: ubuntu-latest
     strategy:
+      fail-fast: false
```
bpkroth, line 89: TODO: consider removing the `fail-fast: false` rule for each of these.
On the one hand it makes failing PRs fail quicker.
On the other hand it means going through several (long) iterations before we can tell what all is broken.

bpkroth:
Currently also has some changes from #277 included.  Waiting for that to merge first.

bpkroth:
Just a few more failures in sqlite left to fix ...

bpkroth:
@anjagruenheid @jcamachor if either of you have a moment, could you please give this a once over?  Thanks!

src/main/java/com/oltpbenchmark/benchmarks/tpch/procedures/GenericQuery.java
```diff
@@ -34,12 +35,18 @@ public abstract class GenericQuery extends Procedure {
     protected abstract PreparedStatement getStatement(Connection conn, RandomGenerator rand, double scaleFactor) throws SQLException;
 
     public void run(Connection conn, RandomGenerator rand, double scaleFactor) throws SQLException {
-
-        try (PreparedStatement stmt = getStatement(conn, rand, scaleFactor); ResultSet rs = stmt.executeQuery()) {
-            while (rs.next()) {
-                //do nothing
+        try (PreparedStatement stmt = getStatement(conn, rand, scaleFactor)) {
+            try (ResultSet rs = stmt.executeQuery()) {
```
bpkroth, line 39: Splitting these made it slightly easier to see which line the exception was coming from, but it's probably not necessary.

src/main/java/com/oltpbenchmark/benchmarks/wikipedia/WikipediaLoader.java
```diff
@@ -101,6 +101,8 @@ public List<LoaderThread> createLoaderThreads() {
             public void load(Connection conn) throws SQLException {
                 Table catalog_tbl = benchmark.getCatalog().getTable(WikipediaConstants.TABLENAME_USER);
 
+                SQLUtil.setIdentityInsert(conn, getDatabaseType(), catalog_tbl, true);
```
bpkroth, line 104: Basically a SqlServer specific workaround for being able to insert into a table with auto-increment columns.

src/main/resources/benchmarks/tpcc/dialect-phoenix.xml
bpkroth, line 1: These files were mistakenly named and not getting picked up.  Renaming them to the expected format fixes those issues.

src/main/resources/benchmarks/tpch/dialect-mariadb.xml
bpkroth, line 1: A copy of the hsqldb dialect file to address some `BETWEEN date '?'` syntax that differs slightly between mariadb and mysqldb.

src/main/java/com/oltpbenchmark/types/DatabaseType.java
```diff
@@ -45,16 +45,21 @@ public enum DatabaseType {
     ORACLE(true, false),
     SINGLESTORE(true, false),
     SPANNER(false, true),
-    SQLAZURE(true, false),
+    SQLAZURE(true, true, true),
     SQLITE(true, false),
-    SQLSERVER(true, false),
+    SQLSERVER(true, true, true),
```
bpkroth, line 50: Sqlserver needs column names for the `INSERT IDENTITY` setting to apply.

src/main/java/com/oltpbenchmark/types/DatabaseType.java
```diff
@@ -45,16 +45,21 @@ public enum DatabaseType {
...
     TIMESTEN(true, false),
     PHOENIX(true, true);
 
 
-    DatabaseType(boolean escapeNames, boolean includeColNames) {
+    DatabaseType(boolean escapeNames, boolean includeColNames, boolean loadNeedsUpdateColumnSequence) {
```
bpkroth, line 55: Per PR feedback, we use this additional DbType feature to consolidate the check for whether we should attempt to update column sequences after loading data.

.github/workflows/maven.yml
bpkroth: Leaving this as is for now.

PR #277, opened 2023-03-31
Dev build cycle improvements

bpkroth:
- Adds more assembly descriptors for producing flat directories to avoid having to compress/extract when building containers.
- Provides a way to selectively enable that during container builds.
- Skips generating the zip files during the CI pipelines (they were unused anyways).
- Adds some unrelated minor fixups to more docker-compose scripts following #276


src/main/assembly/dir.xml
```diff
@@ -0,0 +1,48 @@
+<assembly xmlns="http://maven.apache.org/ASSEMBLY/2.0.0"
+          xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+          xsi:schemaLocation="http://maven.apache.org/ASSEMBLY/2.0.0 http://maven.apache.org/xsd/assembly-2.0.0.xsd">
+    <id>dir</id>
+    <includeBaseDirectory>true</includeBaseDirectory>
+
+    <formats>
+        <format>dir</format>
```
bpkroth, line 8: Adds a copy of the `tgz` version of this file for just producing a flat directory that can be used when generating docker images.

PR #283, opened 2023-04-04
Connection exception and reconnection handling

### LINKED ISSUE #32: Detect closed connections and automatically attempt to reconnect them.
The Worker creates a connection [here](https://github.com/cmu-db/benchbase/blob/main/src/main/java/com/oltpbenchmark/api/Worker.java#L75) in its constructor.

However, if the connection is closed by the DBMS (e.g., hammer a CPU-limited PostgreSQL read replica with reads to the point of replication replay being blocked), the Worker does not recover and tries to execute on the closed connection.

We would like the closed connection to be reopened. Maybe add a setting for whether this behavior should happen.

bpkroth:
Initial work to provide connection error handling that optionally reconnects.

Fixes #32 

Still probably needs:
- [ ] Unit tests (open to suggestions on how to handle this, but for now skipping in favor of functional tests in the CI pipeline and local script infrastructure)
- [x] Expanded/tested support for additional database engines.

Additional work on improving test infrastructure has been split out to #427.


src/main/java/com/oltpbenchmark/api/Worker.java
```diff
@@ -483,6 +556,42 @@ protected final void doWork(DatabaseType databaseType, TransactionType transacti
...
+     * Checks to see if the exception indicates that the current connection is read-only.
+     * @param ex
+     * @return
+     */
+    private boolean indicatesReadOnly(SQLException ex) {
+
+        String sqlState = ex.getSQLState();
+        int errorCode = ex.getErrorCode();
+
+        LOG.debug("sql state [{}] and error code [{}]", sqlState, errorCode);
```
bpkroth, line 569: - [ ] Wrap in `LOG.isDebugEnabled` check

src/main/java/com/oltpbenchmark/api/Worker.java
bpkroth, line 1: This is still pretty messy and could use some cleanup.

src/main/java/com/oltpbenchmark/api/Worker.java
timveil: i thought best practice these days was to forgo the isDebugEnabled type checks.   i'm fine for you to leave that out

src/main/java/com/oltpbenchmark/api/Worker.java
timveil: yeah all this logic in the `catch` block is hard to read.  I'd be in favor of extracting it out to a private method which may help readability.  otherwise looks good.

timveil:
I think these changes look fine.  would like to make sure they work across other db platforms so making this approach extensible is important.

timveil (approved):
looks good as a start.  curious how well this works with other db platforms but we can address as issues are found

src/main/java/com/oltpbenchmark/api/Worker.java
bpkroth: Yeah, I guess for Java it doesn't matter here.  In other places we want to avoid the expensive string formatting operations in case the call isn't necessary.

.github/workflows/maven.yml
```diff
@@ -191,6 +194,10 @@ jobs:
           if [[ ${{matrix.benchmark}} == templated ]]; then
             java -jar benchbase.jar -b tpcc -c config/mariadb/sample_tpcc_config.xml --create=true --load=true --execute=false --json-histograms results/histograms.json
             java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/mariadb/sample_${{matrix.benchmark}}_config.xml --create=false --load=false --execute=true --json-histograms results/histograms.json
+          elif [[ ${{matrix.benchmark}} == tpcc-with-reconnects ]]; then
+            java -jar benchbase.jar -b tpcc -c config/mariadb/sample_tpcc_config.xml --create=true --load=true
+            ./scripts/interrupt-docker-db-service.sh mariadb
```
bpkroth, line 199: Service name detection needs improvement in Github environment

PR #300, opened 2023-04-10
Enable advanced monitoring for SQLServer & Postgres.

anjagruenheid:
This PR introduces advanced monitoring (per query and system stats) for SQL Server and Postgres. Anyone currently using monitoring will not be impacted as advanced monitoring will need to be enabled with a new CL option (mt=advanced).

In essence, this code addition allows users to run queries against system tables (SQL Server) or plugins (Postgres) which allows us to extract system and query usage statistics while benchmarking (query plans, query text, pattern execution counts, cache hits etc.).


src/main/java/com/oltpbenchmark/api/collectors/monitoring/DatabaseMonitor.java
bpkroth, line 1: Possibly this file could use some unit tests on the expected file rotation behavior and whatnot.
Could also mark it as future work.

src/main/resources/benchmarks/tpch/dialect-postgres.xml
```diff
@@ -3,65 +3,76 @@
     <dialect type="POSTGRES">
         <procedure name="Q1">
             <statement name="query_stmt">
+                /*monitor-Q01*/
```
bpkroth, line 6: I wonder if these tweaks could be generalized so that they could automatically be added during dialect construction somehow.
It'd make it applicable beyond this particular workload as well.
For instance, what if each statement had `with_monitoring_comment="true"` property that defaulted to `"true"` if omitted.
Similarly, in the `DatabaseType` enum, each db type could have a `generate_monitoring_comments=true` default, that could be overridden to `false` if that particular DB doesn't accept `/* monitoring-NAME */` style syntax.
Finally, in `Procedure.java` modify `getStatements()` and `setSql()` in `SQLStmt.java` somehow to adjust the query dynamically to include that `/* monitoring-{name} */` comment in the query text?

I feel like this might be more maintainable and generic for the long run.  Thoughts?

bpkroth: BTW, this is a guess on how that implementation might work.  There are various layers of indirection here I haven't totally followed thru yet.

bpkroth: And such a system would also need some tests.

src/main/java/com/oltpbenchmark/api/collectors/monitoring/README.md
```diff
@@ -0,0 +1,19 @@
...
+The former is the default setting unless the monitoring type is explicitly set to advanced which will trigger system monitoring if the database type is supported.
+
+Throughput monitoring logs uptdated throughput values directly to the system output, while advanced monitoring creates csv files recording their findings in folder results/monitor/.
+Advanced monitoring collects data for a variety of events, such as one-off information about a query (for example query plans, query text, etc.), repeated information about a query (elapsed time per query execution, worker time, execution count etc.), and repeated system information (cache hits, number of transactions etc.).
+Which events are collected depends on the database system and is customized in corresponding drivers.
+The code for the drivers can be found in package src.main.java.com.oltpbenchmark.api.collectors.monitoring.
+
+For advanced monitoring to function with SQLServer, the user needs to have access to the system tables, for Postgres, pg_stat_statements needs to be enabled.
```
bpkroth, line 17:
```suggestion
For advanced monitoring to function with SQLServer, the user needs to have access to the system tables, for Postgres, `pg_stat_statements` needs to be enabled.
```
Style nit: code, paths, commands, and other system type info might be easier to read if it was marked as such.

src/main/java/com/oltpbenchmark/api/collectors/monitoring/SQLServerMonitor.java
```diff
@@ -0,0 +1,237 @@
...
+
+/**
+ * Implementation of a monitor specific to SQLServer. Uses SQLServer's system
+ * tables to extract relevant query and system information.
+ */
+public class SQLServerMonitor extends DatabaseMonitor {
+    private final String MONITORING_IDENTIFIER = "/*monitor-";
+    private final String MONITORING_SPLIT_MARKER = "*/";
+
+    private final String DM_EXEC_QUERY_STATS = "SELECT " +
```
bpkroth, line 29: Switch to 
```java
private String sql = """
SELECT ...
"""
;
```
style here docs?

bpkroth: Makes so copy/paste large query blocks for manual testing is easier

src/main/java/com/oltpbenchmark/ThreadBench.java
```diff
@@ -295,6 +301,20 @@ private Results runRateLimitedMultiPhase() {
             }
         }
 
+        // Stop the monitoring thread separately from cleanup all the workers so
+        // we can ignore errors from these threads (including possible
+        // SQLExceptions), but not the others.
+        try {
+            if (this.monitor != null) {
+                this.monitor.interrupt();
+                this.monitor.join(60000); // wait for 60second for monitor thread to rejoin
```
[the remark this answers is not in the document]
anjagruenheid: I set it as a class variable. There's another part in the code where the worker threads rejoin the same way, should I address that as well?

src/main/resources/benchmarks/tpch/dialect-postgres.xml
anjagruenheid: Hmm, good point, I think this could work. In essence, we need the monitoring comment to filter the cached query plans to only store those that are actually part of the benchmark. I'll play around with it to see what I can come up with.

src/main/java/com/oltpbenchmark/ThreadBench.java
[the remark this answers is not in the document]
bpkroth: Probably

src/main/java/com/oltpbenchmark/api/collectors/monitoring/README.md
```diff
@@ -0,0 +1,19 @@
...
+We currently support two types of monitoring: 1) Basic throughput monitoring to track the progress while executing a benchmark (mt=throughput) and 2) monitoring of query and system properties via system tables for both SQLServer and Postgres (mt=advanced).
+The former is the default setting unless the monitoring type is explicitly set to advanced which will trigger system monitoring if the database type is supported.
+
+Throughput monitoring logs uptdated throughput values directly to the system output, while advanced monitoring creates csv files recording their findings in folder results/monitor/.
+Advanced monitoring collects data for a variety of events, such as one-off information about a query (for example query plans, query text, etc.), repeated information about a query (elapsed time per query execution, worker time, execution count etc.), and repeated system information (cache hits, number of transactions etc.).
+Which events are collected depends on the database system and is customized in corresponding drivers.
+The code for the drivers can be found in package src.main.java.com.oltpbenchmark.api.collectors.monitoring.
```
bpkroth, line 15:
```suggestion
The code for the drivers can be found in package [`src.main.java.com.oltpbenchmark.api.collectors.monitoring`](./../monitoring/).
```

src/main/java/com/oltpbenchmark/api/collectors/monitoring/README.md
```diff
@@ -0,0 +1,19 @@
...
+Monitoring in BenchBase can be enabled using the
+```text
+ -im,--interval-monitor <arg>   Monitoring Interval in milliseconds
+ -mt,--monitor-type <arg>       Type of Monitoring (throughput/advanced)
+```
+command line option when executing BenchBase, where the monitoring interval describes the sleeping period of the thread between recording monitoring information.
+We currently support two types of monitoring: 1) Basic throughput monitoring to track the progress while executing a benchmark (mt=throughput) and 2) monitoring of query and system properties via system tables for both SQLServer and Postgres (mt=advanced).
+The former is the default setting unless the monitoring type is explicitly set to advanced which will trigger system monitoring if the database type is supported.
+
+Throughput monitoring logs uptdated throughput values directly to the system output, while advanced monitoring creates csv files recording their findings in folder results/monitor/.
```
bpkroth, line 12:
```suggestion
Throughput monitoring logs updated throughput values directly to the system output, while advanced monitoring creates csv files recording their findings in folder `results/monitor/`.
```

src/main/java/com/oltpbenchmark/api/collectors/monitoring/README.md
```diff
@@ -0,0 +1,19 @@
+# Monitoring in BenchBase
+
+Monitoring in BenchBase can be enabled using the
+```text
+ -im,--interval-monitor <arg>   Monitoring Interval in milliseconds
+ -mt,--monitor-type <arg>       Type of Monitoring (throughput/advanced)
+```
+command line option when executing BenchBase, where the monitoring interval describes the sleeping period of the thread between recording monitoring information.
+We currently support two types of monitoring: 1) Basic throughput monitoring to track the progress while executing a benchmark (mt=throughput) and 2) monitoring of query and system properties via system tables for both SQLServer and Postgres (mt=advanced).
```
bpkroth, line 9:
```suggestion
We currently support two types of monitoring:

1. Basic throughput monitoring to track the progress while executing a benchmark (`-mt=throughput`), and
2. monitoring of query and system properties via system tables for both SQLServer and Postgres (`-mt=advanced`).
    Support for other engines can also be added.

```

Might also be worth editing this in vscode with the markdown lint check extension.

src/main/resources/benchmarks/tpch/dialect-postgres.xml
anjagruenheid: It took me a bit to figure out the best (non-intrusive and generally applicable) way to do this but I think I got it! Basically, I'm setting a flag in the workload configuration that triggers a statement update similar to the dialect adjustments that we currently have. That way, it's generic and applied for a) any enabled DBMS and b) for any benchmark.

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -81,11 +83,17 @@ public static void main(String[] args) throws Exception {
...
+        ImmutableMonitorInfo.Builder builder = ImmutableMonitorInfo.builder();
         if (argsLine.hasOption("im")) {
-            intervalMonitor = Integer.parseInt(argsLine.getOptionValue("im"));
+            builder.monitoringInterval(Integer.parseInt(argsLine.getOptionValue("im")));
+        }
+        if (argsLine.hasOption("mt")) {
+            if (argsLine.getOptionValue("mt").equals("advanced")) {
+                builder.monitoringType(MonitorInfo.MonitoringType.ADVANCED);
+            }
         }
```
bpkroth, line 95: Maybe restructure this a tad so that if `-mt` is passed and it's not one of the known formats we throw an error about invalid cli args.

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -150,6 +158,13 @@ public static void main(String[] args) throws Exception {
                 // Nothing to do here !
             }
 
+            // Set monitoring enabled, if all requirements are met.
+            if (monitorInfo.getMonitoringInterval()>0 && monitorInfo.getMonitoringType()==MonitoringType.ADVANCED
```
bpkroth, line 162:
```suggestion
            if (monitorInfo.getMonitoringInterval() > 0 && monitorInfo.getMonitoringType() == MonitoringType.ADVANCED
```

src/main/java/com/oltpbenchmark/WorkloadConfiguration.java
```diff
@@ -124,6 +125,14 @@ public void setMaxRetries(int maxRetries) {
         this.maxRetries = maxRetries;
     }
 
+    public void setAdvancedMonitoringEnabled(boolean advancedMonitoringEnabled) {
+        this.advancedMonitoringEnabled = true;
+    }
+
+    public boolean getMonitoringEnabled() {
```
bpkroth, line 132:
```suggestion
    public boolean getAdvancedMonitoringEnabled() {
```

?

src/main/java/com/oltpbenchmark/api/Procedure.java
```diff
@@ -156,6 +158,24 @@ protected final void loadSQLDialect(StatementDialects dialects) {
         }
     }
 
+    /**
+     * Enable monitoring for this procedure by adding a monitoring prefixes.
+     */
+    protected final void enabledAdvancedMonitoring() {
```
bpkroth, line 164:
```suggestion
    protected final void enableAdvancedMonitoring() {
```

src/main/java/com/oltpbenchmark/api/Procedure.java
```diff
@@ -156,6 +158,24 @@ protected final void loadSQLDialect(StatementDialects dialects) {
...
+        for (String stmtName : this.getStatements().keySet()) {
+            SQLStmt stmt = this.name_stmt_xref.get(stmtName);
+            if (LOG.isDebugEnabled()) {
+                LOG.debug(String.format("Enabling advanced monitoring for query %s",
```
bpkroth, line 168:
```suggestion
                LOG.debug(String.format("Enabling advanced monitoring for query %s",
```

@timveil pointed out that the easier way to do this now is with "parameterized logging"
https://www.slf4j.org/faq.html#logging_performance
So then you can also drop the `if (LOG.isDebugEnabled())` since the string parameterization parsing overhead will only take effect when debug is actually enabled.
I think there are cases where the arguments passed to the string format might still be expensive so the `if (LOG.isDebugEnabled())` check might still make sense then, but for others it'll be more readable to use the parameterized format and drop the check.

src/main/java/com/oltpbenchmark/api/collectors/monitoring/PostgreSQLMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+            ResultSet rs = stmt.executeQuery();
+            while (rs.next()) {
+                // Only store those queries that have monitoring enabled via a
+                // comment in the SQL Server dialect XML.
+                String query_text = rs.getString("query_text");
+                if (!query_text.contains(MonitoringUtil.getMonitoringPrefix())) {
+                    continue;
+                }
+                // Get identifier from commment in query text.
+                String[] split = query_text.split(Pattern.quote(MonitoringUtil.getMonitoringPrefix()));
```
bpkroth, line 88: Since you're already using regexes, could you not use a regex capture to pull the string name out and drop the splitting and suffix tracking?

I think the first prefix and the simple `if (query_text.contains(MonitoringUtil.getMonitoringPrefix()))` makes sense as an efficient initial filter if you expect many items not to match, but then after that you could probably do something like this for the queryId extraction:

```java
class MonitoringUtil {
    public static final Pattern monitoringCommentPattern = Pattern.compile("/[*] monitoring-(\S+) */");

    // or even with named captures
    public static final Pattern MonitoringCommentPattern = Pattern.compile("/[*] monitoring-(?<queryId>\S+) */");
    // the exact regex may require tweaking if you don't have the surrounding whitespace in the comment text for instance
```

Now, when you go to match the query text you could do something like this:
```java
Matcher m = MonitoringUtil.MonitoringCommentPattern.matcher(query_text)
if (m.find()) {
    queryId = m.group(1);
    // or if using named capture group
    queryId = m.group("queryId");  // I think this potentially has slightly higher overhead, but I find it more readable.
}
```

src/main/java/com/oltpbenchmark/api/collectors/monitoring/SQLServerMonitor.java
```diff
@@ -0,0 +1,239 @@
...
+        st.min_logical_reads, st.max_logical_reads, st.total_logical_reads, 
+        st.min_used_grant_kb, st.max_used_grant_kb, st.total_used_grant_kb, 
+        st.min_used_threads, st.max_used_threads, st.total_used_threads 
+        FROM sys.dm_exec_query_stats st 
+        CROSS APPLY sys.dm_exec_sql_text(st.plan_handle) q 
+        CROSS APPLY sys.dm_exec_query_plan(st.plan_handle) pl
+        """;
+    private String DM_OS_PERFORMANCE_STATS = """
+        SELECT cntr_value, counter_name FROM sys.dm_os_performance_counters 
+        WHERE instance_name='$DB_INSTANCE';
```
bpkroth, line 46: Opportunity for parameterized queries instead of string replacement?

src/main/java/com/oltpbenchmark/api/collectors/monitoring/SQLServerMonitor.java
```diff
@@ -0,0 +1,239 @@
...
+                // comment in the SQL Server dialect XML.
+                String query_text = rs.getString("query_text");
+                if (!query_text.contains(MonitoringUtil.getMonitoringPrefix())) {
+                    continue;
+                }
+
+                // Get identifier from commment in query text.
+                String[] split = query_text.split(Pattern.quote(MonitoringUtil.getMonitoringPrefix()));
+                split = split[1].split(Pattern.quote(MonitoringUtil.getMonitoringSuffix()));
+                String identifier = split[0];
```
bpkroth, line 146: Same comments as above about regex capture

src/main/java/com/oltpbenchmark/api/collectors/monitoring/PostgreSQLMonitor.java
```diff
@@ -0,0 +1,151 @@
...
+        } catch (SQLException sqlError) {
+            LOG.error("Error when cleaning up cached plans.");
+            LOG.error(sqlError.getMessage());
+        }
+    }
+
+    @Override
+    protected void writeToCSV() {
+        this.writeSingleQueryEventsToCSV();
+        this.writeRepeatedQueryEventsToCSV();
```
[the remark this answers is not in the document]
bpkroth: cleanupCache also looks like a duplicate.

Maybe that could be in the base class and just skip `if (CLEAN_CACHE == null)` (i.e. make the derived classes only have to define the statement text if it's available.)

src/main/java/com/oltpbenchmark/benchmarks/tpcc/procedures/Delivery.java
```diff
@@ -34,53 +34,66 @@ public class Delivery extends TPCCProcedure {
...
-            "   AND NO_W_ID = ? " +
-            " ORDER BY NO_O_ID ASC " +
-            " LIMIT 1");
+        """
+            SELECT NO_O_ID FROM %s
+             WHERE NO_D_ID = ?
+               AND NO_W_ID = ?
+             ORDER BY NO_O_ID ASC
+             LIMIT 1
+        """.formatted(TPCCConstants.TABLENAME_NEWORDER));
```
bpkroth, line 43: :thumbsup: :heart:

src/main/java/com/oltpbenchmark/types/DatabaseType.java
```diff
@@ -47,19 +47,24 @@ public enum DatabaseType {
...
     SQLAZURE(true, true, true),
     SQLITE(true, false),
-    SQLSERVER(true, true, true),
+    SQLSERVER(true, true, true, true),
     TIMESTEN(true, false),
     PHOENIX(true, true);
 
 
-    DatabaseType(boolean escapeNames, boolean includeColNames, boolean loadNeedsUpdateColumnSequence) {
+    DatabaseType(boolean escapeNames, boolean includeColNames, boolean loadNeedsUpdateColumnSequence, boolean needsMonitoringPrefix) {
```
bpkroth, line 55: Any reason not to default that value to `true` everywhere for now?

src/main/java/com/oltpbenchmark/types/DatabaseType.java
```diff
@@ -47,19 +47,24 @@ public enum DatabaseType {
...
+        this.needsMonitoringPrefix = needsMonitoringPrefix;
+    }
+
+    DatabaseType(boolean escapeNames, boolean includeColNames, boolean loadNeedsUpdateColumnSequence) {
+        this(escapeNames, includeColNames, loadNeedsUpdateColumnSequence, false);
     }
 
     DatabaseType(boolean escapeNames, boolean includeColNames) {
-        this(escapeNames, includeColNames, false);
+        this(escapeNames, includeColNames, false, false);
```
bpkroth, line 67: Any reason not to default that value to `true` everywhere for now?

src/main/java/com/oltpbenchmark/benchmarks/tpcc/procedures/Delivery.java
bpkroth: Strictly speaking these aren't related, right?  So could split this out to a separate PR, but I won't force it ;)

anjagruenheid: Yeah, I was playing around with it separately. Let me split it out, it's the right thing to do.

src/main/java/com/oltpbenchmark/types/DatabaseType.java
anjagruenheid: I think I'd prefer enabling that on a DBMS-by-DBMS basis, whenever the functionality has actually been implemented and tested. I think that makes more sense to me.

.github/workflows/maven.yml
```diff
@@ -422,12 +422,16 @@ jobs:
...
         # Note: user/pass should match those used in sample configs.
         run: |
-          java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/sqlserver/sample_${{matrix.benchmark}}_config.xml --create=true --load=true --execute=true --json-histograms results/histograms.json
+          java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/sqlserver/sample_${{matrix.benchmark}}_config.xml -im 1000 -mt advanced --create=true --load=true --execute=true --json-histograms results/histograms.json
           # FIXME: Reduce the error rate so we don't need these overrides.
           if [ ${{matrix.benchmark}} == tatp ]; then
               ERRORS_THRESHOLD=0.05
           fi
           ./scripts/check_histogram_results.sh results/histograms.json $ERRORS_THRESHOLD
+          if ![ -d "./results/monitor" ]; then
```
anjagruenheid, line 431: @bpkroth : Is this what you had in mind?

src/main/java/com/oltpbenchmark/types/DatabaseType.java
bpkroth: Fair.  I tend to default to yes and let the CI catch errors and then disable it.  That way I at least get a warning that I need to implement something else when adding a new engine for instance.

.github/workflows/maven.yml
bpkroth: Close.  That's just checking that the directory was created though, not that the contents of it were written somewhat sanely.

Also, I think we should enable this check for both Postgres and SqlServer if they're both supported.

anjagruenheid: Replaced with a count for the number of files which should be at least 3 in the current setup (single + repeated query events, system events).

For Postgres to use advanced monitoring, you'd need to deploy an extension first. Is that possible in this pipeline?

anjagruenheid: Also: Any idea why the docker test is failing? I didn't touch it and the error message is not helpful.

bpkroth:
@ranaalotaibiMS here's the PR I was mentioning.  Could potentially extend this to optionally support plan capture as well.

.github/workflows/maven.yml
```diff
@@ -423,13 +423,19 @@ jobs:
       - name: Run benchmark
         # Note: user/pass should match those used in sample configs.
         run: |
-          java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/sqlserver/sample_${{matrix.benchmark}}_config.xml --create=true --load=true --execute=true --json-histograms results/histograms.json
+          java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/sqlserver/sample_${{matrix.benchmark}}_config.xml -im 1000 -mt advanced --create=true --load=true --execute=true --json-histograms results/histograms.json
```
bpkroth, line 426: Extend to postgres as well?

.github/workflows/maven.yml
```diff
@@ -423,13 +423,19 @@ jobs:
...
           # FIXME: Reduce the error rate so we don't need these overrides.
           if [ ${{matrix.benchmark}} == tatp ]; then
               ERRORS_THRESHOLD=0.05
           fi
           ./scripts/check_histogram_results.sh results/histograms.json $ERRORS_THRESHOLD
 
+          # Running the monitor should create at least three files in the 'monitor' directory.
+          if ![ $(find "./results/monitor" -maxdepth 1 -mindepth 1 | wc -l) -gt 2]; then
+            echo "ERROR: Advanced monitoring unsuccessful, file directory and/or appropriate files not created." >&2
+            exit 1
```
bpkroth, line 436: Extend to postgres as well?

.github/workflows/maven.yml
bpkroth: > For Postgres to use advanced monitoring, you'd need to deploy an extension first. Is that possible in this pipeline?

Probably - can you point me at the extension?  I can probably help with that.

.github/workflows/maven.yml
anjagruenheid: Postgres needs a plugin installed before this works, is that even possible?

.github/workflows/maven.yml
anjagruenheid: s.a.

src/main/java/com/oltpbenchmark/api/collectors/monitoring/DatabaseMonitor.java
anjagruenheid: Not sure I understand this comment. There's no unit tests right now, writing to file is tested as part of the pipeline but the content is not checked currently.

bpkroth: End to end test works for me too.

.github/workflows/maven.yml
bpkroth: It should be.  Can you point me at the extension we need?  Maybe include it in some documentation with the PR too?

anjagruenheid: I had added a reference in the README that's in the monitoring folder. This is the name: pg_stat_statements
Let me know if/how enabling it is possible and I'll add the appropriate tests.

PR #301, opened 2023-04-20
Introduce templated benchmarks.

anjagruenheid:
This PR introduces templated benchmarks for BenchBase. In essence, users can define their own query templates that are parametrized with one or more values at runtime. If multiple parameter combinations are possible, we will continue to cycle through all available combinations.

A templated benchmark has the structure
```xml
<templates> 
    <template name="$QueryTemplateName">   
        <query>$SQLQuery</query>   
        <types>
            <type>$ParameterType1</type>
            <type>$ParameterType2</type>
        </types>   
        <values>     
            <value>$ParameterValueA1</value>
            <value>$ParameterValueA2</value>
            ...
        </values>
        <values>     
            <value>$ParameterValueB1</value>
            <value>$ParameterValueB2</value>
            ...
        </values> 
    </template> ...
<templates>
```
where $ParameterType is the integer java.sql.Types value (i.e., 4 for integer, 16 for boolean etc.) and each value tag within 'parameter_values' contains the values for one instantiation of the parameters set in $SQLQuery. The SQL query string is read as a PreparedStatement, i.e., parameters are defined in the string via a '?' placeholder. An example for a templated benchmark can be found in data/templated/example.xml. The file path for the XML template has to be defined in the workload configuration using the 'query_templates_file' tag. An example configuration can be found in config/sqlserver/sample_template_config.xml. The example can be executed if a loaded TPC-C instance is used as JDBC endpoint.
 
Templated benchmarks are instantiated using 'templated' as benchmark class when running BenchBase via the command line.


data/templated/example.xml
```diff
@@ -0,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<query_templates>
+   <query_template>
+      <name>GetOrder</name>
+      <query><![CDATA[SELECT NO_O_ID FROM new_order WHERE NO_D_ID = ? OR NO_D_ID = ? ORDER BY NO_O_ID ASC]]></query>
+      <parameter_types>4,4</parameter_types>
```
bpkroth, line 6: Are the string names for these unambiguous enough that we could use them instead of integers?
I find that to be somewhat harder to read.

bpkroth, line 6:
```suggestion
      <parameter_types>
          <parameter_type>Integer</parameter_type>
          <parameter_type>Integer</parameter_type>
      </parameter_types>
```
This format might also make more sense.

bpkroth: And then something like this for `<parameter_values>`:

```xml
    <query_values>
        <parameter_values>
            <parameter_value>3</parameter_value>
            <parameter_value>4</parameter_value>
        </parameter_values>
        <parameter_values>
            <parameter_value>5</parameter_value>
            <parameter_value>6</parameter_value>
        </parameter_values>
    </query_values>
```
More verbose, but also more natural XML object parsing.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,287 @@
...
+ * defined in the string via a '?' placeholder. An example for a templated
+ * benchmark can be found in data/templated/example.xml. The file path for the 
+ * XML template has to be defined in the workload configuration using the
+ * 'query_templates_file' tag. An example configuration can be found in
+ * config/sqlserver/sample_template_config.xml. The example can be executed if a
+ * loaded TPC-C instance is used as JDBC endpoint.
+ * 
+ * Templated benchmarks are instantiated using 'templated' as benchmark class
+ * when running BenchBase via the command line.
+ */
```
bpkroth, line 84: Maybe move this into a `README.md` in the directory along with some links to examples and just leave a comment here to read the README?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,287 @@
...
+                                + " getQueryTemplateInfo() {\n"
+                                + "        return ImmutableQueryTemplateInfo.builder()\n"
+                                + "          .query(new " + SQLStmt.class.getCanonicalName() + "(\n"
+                                + "             \"" + StringEscapeUtils.escapeJava(qt.getQuery()) + "\"))\n"
+                                + "          .paramsTypes(new int[] {" + qt.getParamsTypes() + "})\n"
+                                + "          .paramsValues(new String[] {"
+                                + getParamsValuesString(qt.getParamsValues()) + "})\n"
+                                + "          .build();\n"
+                                + "    }\n"
+                                + "}";
```
bpkroth, line 220:
```java
s = """
long
string
syntax
instead?
""'
// maybe also with .format() substitution
```
https://www.demo2s.com/java/java-17-string-block-for-multiline-java-string.html

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TraceTransactionGenerator.java
```diff
@@ -0,0 +1,47 @@
...
+        this.nextInTrace = new CyclicCounterGenerator(transactions.size());
+    }
+
+    @Override
+    public GenericQueryOperation nextTransaction() {
+        return transactions.get(nextInTrace.nextInt());
+    }
+
+    public boolean isEmpty() {
+        return transactions.size()==0;
```
bpkroth, line 44:
```suggestion
        return transactions.size() == 0;
```

nit (though we might want to add some sort of style checker to this code base for consistency and to ease work for maintainers and developers)

src/main/java/com/oltpbenchmark/benchmarks/templated/util/GenericQueryOperation.java
```diff
@@ -0,0 +1,41 @@
...
+ * Immutable class containing information about transactions.
+ */
+public class GenericQueryOperation extends Operation {
+
+    public final List<Object> params;
+
+
+    public GenericQueryOperation(Object[] params) {
+        super();
+        this.params =  Collections.unmodifiableList(Arrays.asList(params));
```
bpkroth, line 35:
```suggestion
        this.params = Collections.unmodifiableList(Arrays.asList(params));
```

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedWorker.java
```diff
@@ -0,0 +1,67 @@
...
+            if (!generators.get(clazz).isEmpty()) {
+                // If there is a generator available use it to create a
+                // parameter binding.
+                TraceTransactionGenerator generator = generators.get(clazz);
+                proc.run(conn, generator.nextTransaction().getParams());
+            } else {
+                // If the generator has no transactions, there are no parameters.
+                proc.run(conn);
+            }
+
```
bpkroth, line 57:
```suggestion
```

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,287 @@
...
+                            break;
+                    }
+                }
+
+                // Create a compiled query that contains all parsed information.
+                if (nextEvent.isEndElement()) {
+                    EndElement endElement = nextEvent.asEndElement();
+                    if (endElement.getName().getLocalPart().equals("query_template")) {
+                        ParsedQueryTemplate qt = b.build();
+                        // Create and compile class.
```
bpkroth, line 205: This is an interesting technique 👍

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,287 @@
...
+
+    @Override
+    protected Loader<TemplatedBenchmark> makeLoaderImpl() {
+        // Loader not currently supported.
+        throw new UnsupportedOperationException();
+    }
+
+    private ClassLoader loadQueryTemplates(String file) {
+        // Instantiate Java compiler.
+        CustomClassLoader ccloader = new CustomClassLoader(this.classLoader);
```
bpkroth, line 159: Thought: If you included an XML schema file for the templates could you make use of that to auto load a class from the XML?
If nothing else, it might be good for data validation prior to parsing the content into an executable class.

config/sqlserver/sample_templated_config.xml
bpkroth, line 1: Add a few more of these for other databases as well? (ideally all of them that are supported in the CI pipeline)

data/templated/example.xml
anjagruenheid: Done! It now parses the String from the sql.Types into an integer internally.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
anjagruenheid: Not sure I understand this comment, could you elaborate?

config/sqlserver/sample_templated_config.xml
anjagruenheid: I think all of them should be supported as long as the parameter types are set to values that are supported by the executed DBMS and the queries are parsable by that DBMS. We're not using anything specific to SQL Server in the procedure but use generic PreparedStatements and setting object parameters via the API functions. So as long as the drivers are implemented in that way, it should work imo.

bpkroth: Yeah, I think I agree, so I'm thinking we should add the config examples for them and run them in the CI pipelines too.

https://github.com/cmu-db/benchbase/blob/main/.github/workflows/maven.yml

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
bpkroth: I think this is an example:
https://github.com/cmu-db/benchbase/blob/main/src/main/java/com/oltpbenchmark/api/StatementDialects.java#L128
https://github.com/cmu-db/benchbase/blob/main/src/main/resources/dialect.xsd

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
```diff
@@ -0,0 +1,22 @@
...
+            ...
+        </parameter_values>
+        <parameter_values>     
+            <parameter_value>$ParameterValueB1</parameter_value>
+            <parameter_value>$ParameterValueB2</parameter_value>
+            ...
+        </parameter_values> 
+    </query_template> ...
+<query_templates>
+where $ParameterType is the integer java.sql.Types value (i.e., 4 forinteger, 16 for boolean etc.) and each value tag within 'parameter_values' contains the values for one instantiation of the parameters set in $SQLQuery. The SQL query string is read as a PreparedStatement, i.e., parameters are defined in the string via a '?' placeholder. An example for a templated benchmark can be found in data/templated/example.xml. The file path for the XML template has to be defined in the workload configuration using the'query_templates_file' tag. An example configuration can be found inconfig/sqlserver/sample_template_config.xml. The example can be executed if a loaded TPC-C instance is used as JDBC endpoint.Templated benchmarks are instantiated using 'templated' as benchmark classwhen running BenchBase via the command line.
```
bpkroth, line 22:
```suggestion
where $ParameterType is the `java.sql.Types` value (i.e., Integer, Boolean, etc.) and each value tag within 'parameter_values' contains the values for one instantiation of the parameters set in `$SQLQuery`. The SQL query string is read as a `PreparedStatement`, i.e., parameters are defined in the string via a '?' placeholder. An example for a templated benchmark can be found in data/templated/example.xml. The file path for the XML template has to be defined in the workload configuration using the 'query_templates_file' tag. An example configuration can be found inconfig/sqlserver/sample_template_config.xml. The example can be executed if a loaded TPC-C instance is used as JDBC endpoint. Templated benchmarks are instantiated using 'templated' as benchmark class when running BenchBase via the command line.
```
This comments needs updating right?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
anjagruenheid: Added, but that was no fun!

config/sqlserver/sample_templated_config.xml
anjagruenheid: From what I can tell, the standard process for testing would be setup, load, run, tear down. The problem is that templated benchmarks don't have any 'load' option. You'd have to redefine this differently by creating some sort of external loader which would not conform to the current process.

bpkroth: Should we add support for external create/load scripts too or just have a separate script for that for starters?

src/main/resources/templates.xsd
```diff
@@ -0,0 +1,53 @@
...
+    This schema describes the dialects XML file for a benchmark
+
+    If you modify this schema, you must regenerate the jaxb
+    bindings. It's easy to do:
+
+    cd src/
+    xjc -p com.oltpbenchmark.api.templates com/oltpbenchmark/api/templates.xsd
+
+    You can check that changes to this file produce a valid schema using
+    the public tool at http://www.w3.org/2001/03/webdata/xsv.
```
bpkroth, line 12: Sounds like a good commit check to have.

src/main/resources/templates.xsd
```diff
@@ -0,0 +1,53 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+    This schema describes the dialects XML file for a benchmark
+
+    If you modify this schema, you must regenerate the jaxb
+    bindings. It's easy to do:
+
+    cd src/
+    xjc -p com.oltpbenchmark.api.templates com/oltpbenchmark/api/templates.xsd
```
bpkroth, line 9: Can we automate this too?  e.g. run it during the CI pipelines and make sure that no file changes are generated (e.g. with a `git diff --porcelain` check and abort the checkin since local changes are out of sync.

bpkroth: Other option would be to always autogenerate those as a part of the build and avoid checking them into the repo at all.

anjagruenheid: I think I copied this over from com/oltpbenchmark/api/dialect.xsd which is what you wanted me to mimic. I don't think that's part of the CI pipeline right now and I'm not sure we could auto-generate it as part of the pipeline.

src/main/resources/templates.xsd
anjagruenheid: see comment above

src/main/resources/templates.xsd
bpkroth: Fair.  We can probably leave that for a separate improvement.

config/sqlserver/sample_templated_config.xml
anjagruenheid: It's a bit hacky but I'm loading tpcc and then running the templated benchmark on top of that. What do you think?

.github/workflows/maven.yml
```diff
@@ -423,7 +423,14 @@ jobs:
       - name: Run benchmark
         # Note: user/pass should match those used in sample configs.
         run: |
-          java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/sqlserver/sample_${{matrix.benchmark}}_config.xml --create=true --load=true --execute=true --json-histograms results/histograms.json
+          # For templated benchmarks, we need to preload tpcc data for the test since by design, templated benchmarks do not support the 'load' operation
```
bpkroth, line 426:
```suggestion
          # For templated benchmarks, we need to preload some data for the test since by design, templated benchmarks do not support the 'load' operation
          # In this case, we load the tpcc data.
```

Is this an accurate adjustment?

data/templated/example.xml
```diff
@@ -0,0 +1,36 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bpkroth, line 1:
```suggestion
<?xml version="1.0" encoding="UTF-8"?>
<!-- Note: these example templates reuse TPC-C queries and data as a demonstration, but others are possible. -->
```

config/sqlserver/sample_templated_config.xml
```diff
@@ -0,0 +1,37 @@
...
+    <!-- Connection details -->
+    <type>sqlserver</type>
+    <driver>com.microsoft.sqlserver.jdbc.SQLServerDriver</driver>
+    <url>jdbc:sqlserver://localhost:1433;encrypt=false;database=benchbase;</url>
+    <username>benchuser01</username>
+    <password>P@ssw0rd</password>
+    <isolation>TRANSACTION_SERIALIZABLE</isolation>
+    <batchsize>128</batchsize>
+
+    <query_templates_file>data/templated/example.xml</query_templates_file>
```
bpkroth, line 13:
```suggestion
    <!--
      Note: these example templates reuse TPC-C queries and data as a demonstration, but others are possible. 
      To test this sample as is, you must first run benchbase with `--load=true --config=sample_tpcc_config.xml`
      in order to load the TPC-C data to the target database.
      For others data loading is left as a manual process.
    -->
    <query_templates_file>data/templated/example.xml</query_templates_file>
```

.github/workflows/maven.yml
anjagruenheid: Yep!

config/sqlserver/sample_templated_config.xml
bpkroth: I think it's an OK start for now.  I added some suggestions on documenting that aspect of it a little bit better.  Maybe we can add some TODO comments somewhere for supporting it in the future or some such.

bpkroth: Probably also make sure to disallow that option in the CLI options.

bpkroth: But the other part of this comment was to see whether or not we could make this work for another DBMS as well?  For instance, how possible is it to add it to the Postgres and/or MySQL pipelines as well?
Basically, if there's more than one engine that it should support, I think we should have tests for them to make sure it remains supported.
If others want to get added in the future I think we can add them later as long as it's clearly documented that that's the case.

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
bpkroth, line 1: I left a PR against yours to clean this up a tad:
https://github.com/anjagruenheid/benchbase/pull/1

src/main/resources/templates.xsd
bpkroth: Reopening so I remember to make a new Issue for this.

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
anjagruenheid: Merged! Thank you!

config/sqlserver/sample_templated_config.xml
anjagruenheid: Added and fixed your PR, also added the MySQL pipeline part. Thanks for adding the documentation!

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,251 @@
...
+
+        try {
+            final Map<Class<? extends Procedure>, TraceTransactionGenerator> generators = new HashMap<>();
+            // Create potential parameter bindings for each template. Add those
+            // to a trace transaction generator that will determine how the
+            // parameters are used.
+            for (Entry<TransactionType, Procedure> kv : getProcedures().entrySet()) {
+                // Sanity check that the procedure has the right type.
+                if (!(kv.getValue() instanceof GenericQuery)) {
+                    continue;
```
bpkroth, line 100: Maybe log an error here instead?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,251 @@
...
+                }
+                GenericQuery proc = (GenericQuery) kv.getValue();
+                QueryTemplateInfo info = proc.getQueryTemplateInfo();
+
+                // Parse parameter values and add each combination to a generator.
+                List<GenericQueryOperation> list = new ArrayList<>();
+                String[] paramsTypes = info.getParamsTypes();
+                for (String binding : info.getParamsValues()) {
+                    CSVParser parser = new CSVParserBuilder()
```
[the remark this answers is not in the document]
bpkroth: Ah, I think I might see the problem.  In the initial loader you're still converting the `<values>` to a comma separated list, and then parsing it back out later, but that has a `, ` instead of just `,` as the separator currently.

I have a fix for that in my branch off of yours plus a few others.

Working on making unit tests work as well so that debugging these are hopefully a little bit easier.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedWorker.java
```diff
@@ -0,0 +1,66 @@
...
+import com.oltpbenchmark.types.TransactionStatus;
+
+public class TemplatedWorker extends Worker<TemplatedBenchmark> {
+
+    protected final Map<Class<? extends Procedure>, TraceTransactionGenerator> generators;
+
+    public TemplatedWorker(TemplatedBenchmark benchmarkModule, int id,
+                     Map<Class<? extends Procedure>, TraceTransactionGenerator> generators) {
+        super(benchmarkModule, id);
+        this.rng().setSeed(15721);
```
bpkroth, line 38: We should probably pull this from the `workConf` instead of statically assigning it.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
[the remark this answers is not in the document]
bpkroth: Another PR for this against your branch here:
https://github.com/anjagruenheid/benchbase/pull/2

There's a few niggling issues with it (lack of NULL tests), but it appears to be passing for me now.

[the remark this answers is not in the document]
anjagruenheid: I like it, should I merge it or do you want to do more edits?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedWorker.java
anjagruenheid: The seed? I think I copied this from the TPC-H worker which has the same setup.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
anjagruenheid: done!

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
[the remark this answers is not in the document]
bpkroth: If you agree, I think you can merge it as is.  I added a few TODOs in it to expand some test coverage (mostly NULL handling), but I think you could handle that too, right?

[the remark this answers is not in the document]
bpkroth: Actually, I fixed a few more things in it and I think it should be fine as is.
Let's take up the lack of support for NULLable params in a future PR and merge this one as is.
Sound good?
I can create a new issue for tracking that.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedWorker.java
bpkroth: Hmm, I wonder if that one is part of the TPC-H spec?
I think we should let it be config controllable here though.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -0,0 +1,255 @@
...
+                // Create and compile class.
+                final String s = """
+                        package %s ;
+                        public final class %s extends %s {
+                            @Override
+                            public %s getQueryTemplateInfo() {
+                                return ImmutableQueryTemplateInfo.builder()
+                                        .query(new %s(\"%s\"))
+                                        .paramsTypes(new String[] {%s})
+                                        .paramsValues(new String[] {%s})
```
bpkroth, line 178: For future NULLable handling, I wonder if it might be possible to change this to pass a multi-dimensional array of objects from the XML parser so that the CSV re-parsing portion can be skipped.
That would be a step towards supporting passing NULLs in, but I think the XML schema might also have to be adjusted somehow to account for it - right now the values received are `""` even when using something like `<value />`

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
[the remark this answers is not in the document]
anjagruenheid: Yep, sounds good!

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
anjagruenheid: Right now we're reading the input as a String[], then use the CSV Parser to parse it into an Object[]. The actual type is not set until the query is triggered, using the SQL.Type to set the object via the JDBC driver. There's two pain points: The instantiation into a class which we currently do as String[] to avoid any object-specific handling and then the conversion into the Object[]. The CSV Parser was initially designed to help with parsing different types of objects, if memory serves, but I agree that it's convoluted.

.github/workflows/maven.yml
```diff
@@ -112,7 +112,14 @@ jobs:
 
       - name: Run benchmark
         run: |
-          java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/sqlite/sample_${{matrix.benchmark}}_config.xml --create=true --load=true --execute=true --json-histograms results/histograms.json
+          # For templated benchmarks, we need to preload some data for the test since by design, templated benchmarks do not support the 'load' operation
+          # In this case, we load the tpcc data.
+          if [[ ${{matrix.benchmark}} == templated ]]; then
```
bpkroth, line 117: Looks like I missed listing the benchmark in the matrix above.  Can you add that too please?

config/mariadb/sample_templated_config.xml
```diff
@@ -0,0 +1,49 @@
...
+      Other templated benchmarks need to ensure that the data is loaded manually prior to running the benchmark.
+    -->
+
+    <query_templates_file>data/templated/example.xml</query_templates_file>
+
+    <!-- The workload -->
+    <terminals>1</terminals>
+    <works>
+        <work>
+            <time>10</time>
```
bpkroth, line 25: Also, maybe for CI/CD pipeline runs we don't need to run this quite so long.  Reduce to 5 minutes perhaps?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
bpkroth: Good notes for the future.  Thanks.  Hopefully whoever takes this issue (#366) up later will read these threads again :)

.github/workflows/maven.yml
bpkroth: Trying to see if I can figure out why the docker build was failing next ...

bpkroth: Here's the full set of fixes I applied (skipped a PR this time, hope that's OK):
https://github.com/bpkroth/benchbase/compare/681c0aa7..bc96fcad

Basically, there's 
- a few more fixes to the local testing scripts
- an tweak to speed up tpcc data loading (quicker testing)
- a fix to a prior change I made while enabling unit test support for templated benchmarks
- a fix for a bug I hadn't encountered previously where the wrong driver is used to double check the catalog for sqlite
- a fix to the LIMIT syntax used in a test template query (sqlserver uses TOP instead)

bpkroth (approved):
I think this is ready to go.
It's got lots of tests, both unit and functional for several major DBs now.
The only remaining issue was support for NULLable fields and we filed #366 to track that in future work.

PR #328, opened 2023-06-26
MariaDB CI Fixups

bpkroth:
- Address a change in the MariaDB docker image that replaces `mysqladmin` with `mariadb-admin` that causes our health check command for the CI pipeline service setup to fail.
  - Additionally note a bug in their code that causes a non-zero exit upon auth failures.
- Semi related: ensure that `openssh-client` is included in the devcontainer.
  - Used to be pulled in by `git`, but no longer.


apavlo (approved):
LGTM

PR #331, opened 2023-06-28
Docker script tweaks

bpkroth:
- Support testing server versions other that `latest` with docker compose scripts.
  - Fixup user setup for mysql-5 compose config.
- Fixup some variable unset issues when running outside of the devcontainer.

### DIFF
--- a/docker/benchbase/devcontainer/build-in-container.sh
+++ b/docker/benchbase/devcontainer/build-in-container.sh
@@ -2,6 +2,8 @@
 #
 # A simple script for building one or more profiles (in parallel) inside the container.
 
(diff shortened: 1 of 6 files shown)


docker/benchbase/devcontainer/build-in-container.sh
```diff
@@ -2,6 +2,8 @@
 #
 # A simple script for building one or more profiles (in parallel) inside the container.
 
+# TODO: Convert this to a multi-stage build for better caching.
```
bpkroth, line 5: Made an Issue for this: #332

PR #351, opened 2023-08-31
Exclude CockroachDB's hidden column from a catalog

stonewhitener:
CockroachDB adds a hidden column called `crdb_internal_*` when a hash-sharded primary key is used, that should not be directly used via the catalog.

https://www.cockroachlabs.com/docs/stable/hash-sharded-indexes#create-a-table-with-a-hash-sharded-primary-key

### DIFF
--- a/src/main/java/com/oltpbenchmark/util/SQLUtil.java
+++ b/src/main/java/com/oltpbenchmark/util/SQLUtil.java
@@ -512,14 +512,6 @@ WHERE t.name='%s' AND c.name='%s'
 
         Map<String, Table> tables = new HashMap<>();
 
-        List<String> excludedColumns = new ArrayList<>();
-
-        if (databaseType.equals(DatabaseType.COCKROACHDB)) {
-            // cockroachdb has a hidden column called "ROWID" that should not be directly used via the catalog
-            excludedColumns.add("ROWID");
-        }
-
-
         try (ResultSet table_rs = md.getTables(catalog, schema, null, new String[]{"TABLE"})) {
             while (table_rs.next()) {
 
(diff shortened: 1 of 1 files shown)


src/main/java/com/oltpbenchmark/util/SQLUtil.java
```diff
@@ -512,14 +512,6 @@ private static AbstractCatalog getCatalogDirect(DatabaseType databaseType, Conne
 
         Map<String, Table> tables = new HashMap<>();
 
-        List<String> excludedColumns = new ArrayList<>();
-
-        if (databaseType.equals(DatabaseType.COCKROACHDB)) {
-            // cockroachdb has a hidden column called "ROWID" that should not be directly used via the catalog
-            excludedColumns.add("ROWID");
```
apavlo, line 519: Why didn't you just add `rowid` here?

apavlo (changes requested):

src/main/java/com/oltpbenchmark/util/SQLUtil.java
stonewhitener: This is because the hidden columns to be excluded here must be determined using the `startsWith()` method, not the `equals()` method applied to each element of `excludedColumns`. `excludeColumns` is only used to exclude a hidden column `rowid` for CockroachDB.

stonewhitener: For example, the DDL below generates a hidden column `crdb_internal_ts_shard_20`.

```sql
CREATE TABLE products (
    ts DECIMAL PRIMARY KEY USING HASH WITH (bucket_count = 20),
    product_id INT8
    );
```

A suffix of a hidden column name is determined based on the primary key name and bucket size specified in the DDL. We can not exclude those columns by simply applying `equals()`.

apavlo: Actually, what if we create a list of compiled regular expressions and then have it check that instead having separate logic for `equals` and `startsWith`.

PR #354, opened 2023-09-12
Retry CloseAuctions procedure in AuctionMark on rollback

nuno-faria:
The current way the CloseAuctions procedure is issued is with custom code in the executeWork method, instead of scheduled by the worker. This means that if the transaction fails on concurrency-induced conflicts, when the worker tries to retry, it will retry a different one:
```java
// CloseAuctions conflicting with other transactions 
org.postgresql.util.PSQLException: ERROR: could not serialize access due to read/write dependencies among transactions
  Detail: Reason code: Canceled on identification as a pivot, during conflict out checking.
  Hint: The transaction might succeed if retried.
	at org.postgresql.core.v3.QueryExecutorImpl.receiveErrorResponse(QueryExecutorImpl.java:2713)
	at org.postgresql.core.v3.QueryExecutorImpl.processResults(QueryExecutorImpl.java:2401)
	at org.postgresql.core.v3.QueryExecutorImpl.execute(QueryExecutorImpl.java:368)
	at org.postgresql.jdbc.PgStatement.executeInternal(PgStatement.java:498)
	at org.postgresql.jdbc.PgStatement.execute(PgStatement.java:415)
	at org.postgresql.jdbc.PgPreparedStatement.executeWithFlags(PgPreparedStatement.java:190)
	at org.postgresql.jdbc.PgPreparedStatement.executeUpdate(PgPreparedStatement.java:152)
	at com.oltpbenchmark.benchmarks.auctionmark.procedures.CloseAuctions.run(CloseAuctions.java:168)
	at com.oltpbenchmark.benchmarks.auctionmark.AuctionMarkWorker.executeCloseAuctions(AuctionMarkWorker.java:456)
	at com.oltpbenchmark.benchmarks.auctionmark.AuctionMarkWorker.executeWork(AuctionMarkWorker.java:350)
	at com.oltpbenchmark.api.Worker.doWork(Worker.java:418)
	at com.oltpbenchmark.api.Worker.run(Worker.java:284)
	at java.base/java.lang.Thread.run(Thread.java:833) 

// and when the worker catches the exception, it retries a NewBid instead
Retryable SQLException occurred during [com.oltpbenchmark.benchmarks.auctionmark.procedures.NewBid/03]... current retry attempt [2], max retry attempts [3], sql state [40001], error code [0].
```

Since this procedure is executed sparsely, it would be useful to guarantee that it succeeds.

### DIFF
--- a/src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
+++ b/src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
@@ -447,8 +447,24 @@ public class AuctionMarkWorker extends Worker<AuctionMarkBenchmark> {
         Timestamp startTime = profile.getLastCloseAuctionsTime();
         Timestamp endTime = profile.updateAndGetLastCloseAuctionsTime();
 
-        List<Object[]> results = proc.run(conn, benchmarkTimes, startTime, endTime);
-
+        List<Object[]> results = null;
+        boolean done = false;
+        // Retry here the close auction procedure if it fails due to a transactional conflict, since
+        // the original submitted procedured has a different type.
+        while (!done) {
+            try {
+                results = proc.run(conn, benchmarkTimes, startTime, endTime);
+                done = true;
+            }
+            catch (SQLException e)  {
+                if (e.getSQLState().startsWith("40")) {
+                    conn.rollback();
+                }
+                else {
+                    throw e;
+                }
+            }
+        }
 
         for (Object[] row : results) {
             ItemId itemId = this.processItemRecord(row);


src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
```diff
@@ -447,8 +447,24 @@ protected boolean executeCloseAuctions(Connection conn, CloseAuctions proc) thro
...
+        boolean done = false;
+        // Retry here the close auction procedure if it fails due to a transactional conflict, since
+        // the original submitted procedured has a different type.
+        while (!done) {
+            try {
+                results = proc.run(conn, benchmarkTimes, startTime, endTime);
+                done = true;
+            }
+            catch (SQLException e)  {
+                if (e.getSQLState().startsWith("40")) {
```
apavlo, line 460: What is "40"? Is this DBMS specific?

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
```diff
@@ -447,8 +447,24 @@ protected boolean executeCloseAuctions(Connection conn, CloseAuctions proc) thro
...
+                    conn.rollback();
+                }
```
apavlo, line 462: Can this get stuck in an infinite loop?

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
```diff
@@ -447,8 +447,24 @@ protected boolean executeCloseAuctions(Connection conn, CloseAuctions proc) thro
...
-
+        List<Object[]> results = null;
+        boolean done = false;
+        // Retry here the close auction procedure if it fails due to a transactional conflict, since
+        // the original submitted procedured has a different type.
+        while (!done) {
+            try {
+                results = proc.run(conn, benchmarkTimes, startTime, endTime);
+                done = true;
+            }
```
apavlo, line 458: I actually don't think we want to do this because we can't keep track of the # of times that a txn is submitted. So this can cause BenchBase to exceed the defined submission rate. Also, by retrying inside of the benchmark worker we can't keep track of the # failed txns.

apavlo (commented):
I don't think we want to accept this.

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
nuno-faria: SQL states starting with 40 are related to transactional conflicts, such as "serialization failure" and "integrity constraint violation". As far as I know, it is standard across SQL systems.

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
nuno-faria: I don't think so, because sooner or later the transaction will be able to complete. Any other type of error other than transactional conflicts will raise an exception to the worker.

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkWorker.java
nuno-faria: I also agree that this is not the best approach. I also thought about adding an update method to the "TransactionType" class, so we can update the original "txnType" when we switch it in the "executeWork" function, making the worker deal with retries. This also would mean that the CloseAuctions procedure would be correctly logged in the results, since right now, as far as the worker knows, no CloseAuctions procedure is executed.

PR #356, opened 2023-09-12
Wikipedia benchmark for Oracle DB

ngbanguyen:
Fixes for Wikipedia benchmark to run on Oracle 23c

1. Auto-increment/serial fields can be done with generated as identity in Oracle DB. Removed related sequences and dialect statements
2. In order to run Oracle DB, loading anonymous user in Wikipedia benchmark cannot use empty string as default value for anonymous user, as Oracle treats empty string as null.


apavlo:
@ngbanguyen Why did you close this? We would welcome these fixes

PR #359, opened 2023-09-17
use threadlocal random generator to avoid contention

eivanov89:
Hi,

I noticed a very high CPU consumption, when import initial TPC-C data. According the attached flamegraph all threads are in `TPCCUtil:::randomStr`. Because of util.RandomGenerator there is a contention. This patch uses thread local random generator, which helps to avoid contention and dramatically reduces CPU consumption.

### DIFF
--- a/src/main/java/com/oltpbenchmark/Results.java
+++ b/src/main/java/com/oltpbenchmark/Results.java
@@ -28,6 +28,7 @@ import java.util.Map;
 
 public final class Results {
 
+    private final long startTimestampMs;
     private final long nanoseconds;
     private final int measuredRequests;
     private final DistributionStatistics distributionStatistics;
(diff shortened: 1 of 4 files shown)


apavlo:
@eivanov89 Nice! I will take a look

eivanov89:
@apavlo 
Attached missing flamegraph.

flamegraph-1107381

apavlo (approved):
I am trying to decide if we should just drop `com.oltpbenchmark.util.RandomGenerator` and replace it with `ThreadLocalRandomGenerator`. We can just replace that implementation with the new one (using the same name).

I don't think there will be an non-determinism issue...

eivanov89:
@apavlo ping

apavlo:
@eivanov89 Thanks for poking me. I will look at this in more detail later this week.

apavlo:
@eivanov89 Let's have your RNG class replace the existing one. Can you do this?

eivanov89:
> @eivanov89 Let's have your RNG class replace the existing one. Can you do this?

@apavlo Sure, I can try to do it next week.

eivanov89:
> @eivanov89 Let's have your RNG class replace the existing one. Can you do this?

@apavlo I'm sorry for a long delay. [Here](https://github.com/cmu-db/benchbase/pull/399) is the updated pull request.

PR #361, opened 2023-09-22
Initial draft for Stitcher workload (SQLServer only).

anjagruenheid:
Implementation of Stitcher workload as described in "Stitcher: Learned Workload Synthesis from Historical Performance Footprints.": https://openproceedings.org/2023/conf/edbt/paper-19.pdf by Chengcheng Wan, Yiwen Zhu, Joyce Cahoon, Wenjing Wang, Katherine Lin, Sean Liu, Raymond Truong, Neetu Singh, Alexandra M. Ciortea, Konstantinos Karanasos, Subru Krishnan, published in EDBT 2023.


config/sqlserver/stitcher/2023-02-24-15-11-28-957898.xml
bpkroth, line 1: Would be nicer I think to move these into a subdirectory.
Even better would be to provide a config mechanism that would allow something like the following:

```xml
<params>
   <!-- db specific connect info -->
   <include_workload_info>data/stitcher/workload_a.xml</include_workload_info>
</params>
```

Now have `data/stitcher/workload_a.xml` contain something like the following:

```xml
<works>
   <work>
      <comment>timestamp: 2023-02-24-15-11-28-957898, some other data</comment>
      <time>...</time>
      <rate>...</rate>
      <weights>...</weights>
      <terminals>...</terminals> <!-- moving this parameter in here might require more work -->
   </work>
   <!-- repeat as necessary all in a single file -->
    ...
</works>
```

scripts/stitcher/stitcher_sqlserver.sh
```diff
@@ -0,0 +1,1029 @@
...
+sleep 300
+echo "023-02-02 03:15:00========================================"
+
+sleep 300
+echo "023-02-02 03:20:00========================================"
+
+sleep 300
+echo "023-02-02 03:25:00========================================"
+
+echo "END OF DAY"
```
bpkroth, line 1029:
```suggestion
echo "END OF DAY"

# Wait for all subprocesses to return.
wait
```

scripts/stitcher/README.md
```diff
@@ -0,0 +1,27 @@
+# Stitcher Benchmark
+
+This shell script is used to execute a collection of benchmarks over 24 hours that mimic a resource utilization pattern observed as part of the work done in the following paper:
+[Stitcher: Learned Workload Synthesis from Historical Performance Footprints.](https://openproceedings.org/2023/conf/edbt/paper-19.pdf), Chengcheng Wan, Yiwen Zhu, Joyce Cahoon, Wenjing Wang, Katherine Lin, Sean Liu, Raymond Truong, Neetu Singh, Alexandra M. Ciortea, Konstantinos Karanasos, Subru Krishnan. In EDBT 2023.
+If you use this workload, please reference this paper appropriately.
+
+Prior to executing the shell scripts, the data needs to be preloaded with the following commands:
+
+```sh
+java -jar benchbase.jar -b tpcc -c config/$dbms/stitcher/2023-02-24-15-11-28-957898.xml --create=true --load=true --execute=false
```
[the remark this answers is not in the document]
bpkroth: Also, is it possible for us to have instructions for generating the config sequence given a resource utilization trace?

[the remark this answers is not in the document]
anjagruenheid: I'm not sure how these were originally generated but I don't think the date matters that much? My guess is that these were the configurations of experiments that they collected telemetry for and that were then (partially) picked to mimic this one specific workload. If you look at the configurations, they have different SFs but also different query weight distributions. You would preload 4 datasets (TPC-H, TPC-C SF 16 and 160, and YCSB) and then execute these different configurations on top of those preloaded instances. This benchmark is a static snapshot of a real-world benchmark that we can publish (i.e., add to an open-source repo) because it was already used in published work. Afaik, there is some discussion as to whether they want to open-source the Stitcher code but no resolution yet.

config/sqlserver/stitcher/2023-02-24-15-11-28-957898.xml
anjagruenheid: Okay, thinking about it some more, I think we could generalize this as a new type of benchmark (multi-benchmark?) where the user specifies as input a set of configurations and an execution order as part of its xml definition. The execution order can either contain a pointer to a configuration or a 'sleep' command. The tricky part is rewriting the DBWorkload.java class because you'd need a way to instantiate BenchBase in subthreads. I can give it a go if you're okay with that plan? The stitcher workload would then be one instantiation of such a multi-benchmark.

PR #364, opened 2023-09-26
Support for Sybase ASE + Informix

anasY2:
Added tpcc and ycsb support for sybase ASE and Informix. `pom.xml` has a new profile named `arcion` which will deploy a jar file that contains all the database drivers.


apavlo:
@anasY2 What is the meaning for "arcion"?

apavlo (changes requested):
Thanks for sending this. Can you address the comments? It's not clear why you are bringing in dependencies for Snowflake/MongoDB?

anasY2:
> @anasY2 What is the meaning for "arcion"?

Its a profile name. Instead of downloading drivers separately, it will do it all at once so you do not need to build everytime you wanted to run workload for a separate database.

Issue #360, opened 2023-09-18
Delivery transaction doesn't confirm TPC-C standard


eivanov89:
Hi folks,

At @YDB we also forked benchbase to implement TPC-C. Now we both share same bugs :)

[Here](https://github.com/cmu-db/benchbase/blob/main/src/main/java/com/oltpbenchmark/benchmarks/tpcc/procedures/Delivery.java#L107) you always iterate from 1 to terminalDistrictUpperID. It seems that original approach was to iterate from terminalDistrictLowerID (left unused) to terminalDistrictUpperID. It means that first terminal iterates districts from 1 to 2, next terminal - 1 to 3, ..., 1 to 10. But each terminal must iterate 1 to 10 according the standard (and I double checked TPC-C implementation in CockroachDB).

At YDB bugfix decreased tpmc by 4%. In my opinion, this bug is critical.

apavlo:
@eivanov89 Can you send a PR with your fix?

eivanov89:
@apavlo Sure, I will create the PR soon.

PR #368, opened 2023-09-26
Delivery transaction must use all the districts (#360)

eivanov89:


eivanov89:
@apavlo ping

src/main/java/com/oltpbenchmark/benchmarks/tpcc/procedures/Delivery.java
```diff
@@ -104,7 +104,7 @@ public void run(Connection conn, Random gen, int w_id, int numWarehouses, int te
 
         int[] orderIDs = new int[10];
 
-        for (d_id = 1; d_id <= terminalDistrictUpperID; d_id++) {
+        for (d_id = 1; d_id <= w.getDistPerWhse(); d_id++) {
```
apavlo, line 107: Instead of getting the districts per warehouse from `TPCCWorker`, why not just use `TPCCConfig.configDistPerWhse` since it never changes?

apavlo (commented):
Minor request to just use `TPCCConfig.configDistPerWhse` instead of passing around a new variable.

src/main/java/com/oltpbenchmark/benchmarks/tpcc/procedures/Delivery.java
eivanov89: You're right, thank you! I've updated the pull request. And I'm sorry, that it took so long time.

PR #370, opened 2023-10-01
Docs: README update to help new users

ETHenzlere:
Hey everybody,
I am a new user of benchbase and I found it cumbersome to get everything running smoothly with a lack of documentation. Therefore I have decided to update the README.md slightly, such that new users of benchbase have an easier time getting started. The documentation should now provide enough information now to get started smoothly with benchbase.

My two small additions:
1. Prerequisites section with a direct link to the official Java 17 download page
2. Usage Guide subsection that explains where to change the username and password to connect to DB's

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ BenchBase (formerly [OLTPBench](https://github.com/oltpbenchmark/oltpbench/)) is
 
 **Table of Contents**
 
+- [Prerequisites](#prerequisites)
 - [Quickstart](#quickstart)
 - [Description](#description)
 - [Usage Guide](#usage-guide)
@@ -15,6 +16,13 @@ BenchBase (formerly [OLTPBench](https://github.com/oltpbenchmark/oltpbench/)) is
 - [Citing This Repository](#citing-this-repository)
 
 ---
+
+## Prerequisites
+
+Java 17 needs to be installed on the system in order to run the benchbase commands.
+
+It can be downloaded directly from the [official download page.](https://www.oracle.com/java/technologies/downloads/#java17)
+
 
 ## Quickstart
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -16,6 +17,13 @@ BenchBase (formerly [OLTPBench](https://github.com/oltpbenchmark/oltpbench/)) is
 
 ---
 
+## Prerequisites
+
+Java 17 needs to be installed on the system in order to run the benchbase commands.
+
+It can be downloaded directly from the [official download page.](https://www.oracle.com/java/technologies/downloads/#java17)
```
bpkroth, line 24: This isn't true if you use the `docker` version.

Perhaps split this section out into the following subsections and add it there.
- Containerized
- Non-Containerized

README.md
```diff
@@ -151,6 +159,17 @@ mvn clean compile exec:java -P postgres -Dexec.args="-b tpcc -c config/postgres/
...
 
+### How to configure Database access
+
+The configuration files located in the `/config` folder define the username and password needed to connect to the databases. They can be adapted on a global level by editing the `*_config.xml` files of the corresponding database.
+
+If there is already a built distribution, the config can also be adapted directly in the `/config` folder of the build.
+
+```xml
+<username>admin</username>
+<password>password</password>
```
bpkroth, line 170: IMO, these are the obvious ones and somewhat self explanatory.

We should probably have additional docs on what the other fields in those configs do and how to adjust them.

For instance, it's possible to run on a select set of queries, or multiple workload phases with a single config, but that isn't super obviously documented IMO.

ETHenzlere: I can see that they look obvious to experienced users but they are needed for a quick-start of benchbase so I'd like to keep them here if possible :)

Do you think it would be beneficial to add a new .README file to the config folder and explain the advanced things there? - Or move everything related to config there and add a quick navigation link in the github README?

bpkroth: My point about those two parameters being obvious is that they're self explanatory  _once_ you know where to look, but the part about where the configs are may not be.

I would suggest that the Getting Started text get rewritten as something slightly more high-level like "adjusting the connection details in the config file" which includes basic commentary about both the JDBC URL and the credentials.

A separate `ConfigDetails.md` might be warranted that the main `README.md` links to for additional details, especially with external links to JDBC driver details.

I don't think the entire `ConfigDetails.md` is necessary for you to write here, but if you could start a stub of it with the above, then I think one of the benchbase experts can take up the rest from there.

Make sense?

ETHenzlere:
@bpkroth Hey Brian, I finally found some time to make the changes we talked about. I hope that some of the _benchbase-cracks_ can adapt the `config/README.md` file with more useful information.

PR #379, opened 2023-10-17
Oracle db support

ngbanguyen:


ngbanguyen:
Waiting for approvals from supervisor internally before reopening.

ngbanguyen:
Added Oracle profile as an optional feature. Building and running Oracle profile is similar to other benchmarks.
```bash
./mvnw clean package -P oracle
```
Tested with Oracle Database Express Edition 18.4.0, 21.3.0 and Oracle DB Free Release 23.2.0.0.

Some changes that need explanations:
- General changes to avoid creating empty string when running Oracle profile, as Oracle DB treats empty strings as null.
- `oracle.sql.TIMESTAMP` is not an instance of `java.sql.Timestamp`. `oracle.sql.TIMESTAMP` (and the conversion method)   needs to be reflectively handled as it is OJDBC-specific (not available in JDBC).
- Handles CLOB conversion for LoadConfig procedure in auctionmark and seats. OJDBC CLOB is an instance of `java.sql.Clob` so this can be handled in-code. However, CLOB needs to be read/converted to String while the connection that retrieved it is still open.
- Added an option to run script after loading. `tpch` and `chbenchmark` sample configs contain this option.
  - There were performance issues while running `tpch` and `chbenchmark`. Running `dbms_stats.gather_schema_stats`  would solve the performance issue. The sample Oracle config for chbenchmark and tpch was appended with the option to make sure the optimization was run after loading (especially after running `--create=true --load=true --execute=true`).
    This issue is Oracle profile specific, however it makes more sense to leave this option for other profiles.
- Add and/or amend Oracle DDL and dialects for all benchmarks. I tried to make the changed dialect and/or DDL as close
    to other profiles as possible, mostly referencing MySQL DDL. There are comments for some changes in DDL, such as `resourcestresser` MD5 function in Oracle.
- Official Oracle DB container (container-registry.oracle.com/database/free) doesn't allow creating a user at startup using variables. I chose to use Oracle image from [gvenzl/oci-oracle-xe](https://github.com/gvenzl/oci-oracle-xe/blob/main/README.md) as this image provides a short option for creating a non-DBA user at start up using options.

GitHub Actions with Oracle DB is also added. [Latest run](https://github.com/ngbanguyen/benchbase/actions/runs/6556540941)
  - Using Oracle Database Express Edition 21.3.0 XE.
  - The error rates are higher on some benchmarks due to procedures raising ORA-08177 not serializable.
 
A remaining issue:
- There are some errors while running `auctionmark` on NewPurchase procedure, but I believe the SQL for other procedures are correct. Would appreciate some inputs.

ngbanguyen:
@fgauthie @kavorobyov Please review.

Should I also include [GraalVM Native Image](https://www.graalvm.org/latest/reference-manual/native-image/) configuration file in this PR as well?

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/procedures/LoadConfig.java
```diff
@@ -83,6 +85,12 @@ public Config run(Connection conn) throws SQLException {
         try (PreparedStatement preparedStatement = this.getPreparedStatement(conn, getConfigProfile)) {
             try (ResultSet resultSet = preparedStatement.executeQuery()) {
                 configProfile = SQLUtil.toList(resultSet);
+                // CLOB needs to be done while connection is alive
+                if (dbType == DatabaseType.ORACLE) {
+                    for (Object[] configProfileInstance: configProfile) {
+                        configProfileInstance[3] = SQLUtil.clobToString(configProfileInstance[3]);
+                    }
+                }
```
apavlo, line 93: Can you explain why this is necessary?

src/main/java/com/oltpbenchmark/util/SQLUtil.java
```diff
@@ -109,20 +113,46 @@ public static Double getDouble(Object obj) {
...
+        try {
+            oracleTimestamp = Class.forName("oracle.sql.TIMESTAMP");
+            timestampValueMethod = oracleTimestamp.getDeclaredMethod("timestampValue");
+        } catch (ClassNotFoundException | NoSuchMethodException e) {
+            oracleTimestamp = null;
+            timestampValueMethod = null;
+        }
+        TIMESTAMP_VALUE_METHOD = timestampValueMethod;
+        ORACLE_TIMESTAMP = oracleTimestamp;
+    }
```
apavlo, line 155: What is this code doing? Why is it necessary?

apavlo (changes requested):
Thanks for sending this! Please address the comments in the review.

Most importantly please remove the Oracle copyright. We cannot take this.

.github/workflows/maven.yml
```diff
@@ -58,7 +60,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        profile: [ 'cockroachdb', 'mariadb', 'mysql', 'postgres', 'spanner', 'phoenix', 'sqlserver', 'sqlite' ]
+        profile: [ 'cockroachdb', 'mariadb', 'mysql', 'postgres', 'spanner', 'phoenix', 'sqlserver', 'sqlite' , 'oracle' ]
```
bpkroth, line 63: nit: alphabetize please

.github/workflows/maven.yml
```diff
@@ -271,6 +273,88 @@ jobs:
...
+  oracle:
+    needs: package-and-upload
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        benchmark: [ 'auctionmark', 'epinions', 'hyadapt', 'otmetrics', 'resourcestresser', 'seats', 'sibench', 'smallbank', 'tatp', 'tpcc', 'twitter', 'voter', 'wikipedia', 'ycsb', 'templated' ]
+    services:
+      oracle:
+        image: gvenzl/oracle-xe:21.3.0-slim-faststart
```
bpkroth, line 288: Is this the officially supported Oracle docker image?

docker/oracle/down.sh
```diff
@@ -0,0 +1,8 @@
+#!/bin/bash
+# Copyright (c) 2023,  Oracle and/or its affiliates.
+
+set -eu
+scriptdir=$(dirname "$(readlink -f "$0")")
+cd "$scriptdir/"
+
+docker-compose down --remove-orphans --volumes
```
bpkroth, line 8:
```suggestion
docker compose down --remove-orphans --volumes
```

docker/oracle/reset.sql
bpkroth, line 1: Can this file be used directly in the `.github/workflows/maven.yml` rules instead of the `printf`?

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -455,6 +456,21 @@ public static void main(String[] args) throws Exception {
             LOG.debug("Skipping loading benchmark database records");
         }
 
+        if (xmlConfig.containsKey("afterload")) {
```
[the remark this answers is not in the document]
bpkroth: https://github.com/cmu-db/benchbase/blob/8c86bd461904d46f4c9707e0f88c1b93854504ed/src/main/java/com/oltpbenchmark/api/LoaderThread.java#L65C41-L65C41
That's already in the codebase.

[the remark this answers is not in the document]
bpkroth: However, I do think it'd be good to put that in a separate PR with tests and whatnot.

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
```diff
@@ -248,7 +249,14 @@ public final void createDatabase(DatabaseType dbType, Connection conn) throws SQ
             }
     }
 
-
+    public final void runScript(String scriptPath) throws SQLException, IOException {
```
bpkroth, line 252: As mentioned above - separate PR with individual tests for this I think.

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkLoader.java
```diff
@@ -439,7 +442,8 @@ protected int populateRandomColumns(Object[] row) {
             // STRINGS
             for (Column catalog_col : this.random_str_cols) {
                 int size = catalog_col.getSize();
-                row[catalog_col.getIndex()] = profile.rng.astring(profile.rng.nextInt(size - 1), size);
+                // This can generate an empty string which is treated as NULL in Oracle DB
+                row[catalog_col.getIndex()] = profile.rng.astring(profile.rng.nextInt(getDatabaseType() == ORACLE ? 1 : 0, size - 1), size);
```
bpkroth, line 446:
```suggestion
                init start = getDatabaseType() == ORACLE ? 1 : 0;
                row[catalog_col.getIndex()] = profile.rng.astring(profile.rng.nextInt(start, size - 1), size);
```

bpkroth: For readability

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkProfile.java
bpkroth, line 1: Whitespace and licenses only, please revert.

src/main/java/com/oltpbenchmark/benchmarks/otmetrics/procedures/GetSessionRange.java
```diff
@@ -70,6 +72,7 @@ public List<Object[]> run(Connection conn, int source_id, int session_low, int s
                 finalResults.add(arr);
             }
         }
+        stmt.close();
```
bpkroth, line 75: This isn't Oracle support specific, correct?
Maybe file a separate bug/PR for this?

src/main/java/com/oltpbenchmark/benchmarks/resourcestresser/procedures/CPU1.java
```diff
@@ -36,7 +38,7 @@ public class CPU1 extends Procedure {
         for (int i = 1; i <= ResourceStresserWorker.CPU1_nestedLevel; ++i) {
             complexClause = "md5(concat(" + complexClause + ",?))";
         }
-        cpuSelect = new SQLStmt("SELECT count(*) FROM (SELECT " + complexClause + " FROM " + ResourceStresserConstants.TABLENAME_CPUTABLE + " WHERE empid >= 0 AND empid < 100) AS T1");
+        cpuSelect = new SQLStmt("SELECT count(*) FROM (SELECT " + complexClause + " FROM " + ResourceStresserConstants.TABLENAME_CPUTABLE + " WHERE empid >= 0 AND empid < 100) T1");
```
bpkroth, line 41: Is this cross DBType compatible?

src/main/java/com/oltpbenchmark/util/SQLUtil.java
bpkroth: +1, a few more comments would be nice

src/main/resources/benchmarks/auctionmark/ddl-oracle.sql
```diff
@@ -54,42 +55,42 @@ CREATE TABLE REGION (
...
+    u_sattr1            VARCHAR2(64),
+    u_sattr2            VARCHAR2(64),
+    u_sattr3            VARCHAR2(64),
+    u_sattr4            VARCHAR2(64),
+    u_sattr5            VARCHAR2(64),
+    u_sattr6            VARCHAR2(64),
+    u_sattr7            VARCHAR2(64),
+    u_iattr0            NUMBER(19, 0) DEFAULT NULL,
+    u_iattr1            NUMBER(19, 0) DEFAULT NULL,
+    u_iattr2			NUMBER(19, 0) DEFAULT NULL,
```
bpkroth, line 75:
```suggestion
    u_iattr2            NUMBER(19, 0) DEFAULT NULL,
```
nit: align spacing if you're going to rewrite the file anyways

src/main/resources/benchmarks/chbenchmark/gather_schema_stats_oracle.sql
bpkroth, line 1: Remove license.
Add a comment on what this is doing, how it's invoked, etc., perhaps with some links to documentation.

src/main/resources/benchmarks/otmetrics/ddl-oracle-trimmed.sql
bpkroth, line 1: Is this `-trimmed` version necessary?

src/main/resources/benchmarks/tpch/gather_schema_stats_oracle.sql
```diff
@@ -0,0 +1,2 @@
+-- Copyright (c) 2023, Oracle and/or its affiliates.
+BEGIN dbms_stats.gather_schema_stats(ownname => 'benchbase', estimate_percent => dbms_stats.auto_sample_size, method_opt => 'for all columns size AUTO'); END;;
```
bpkroth, line 2: Looks like this file is repeated.  Possible to consolidate to a single copy?

src/main/java/com/oltpbenchmark/DBWorkload.java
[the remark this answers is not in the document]
ngbanguyen: I needed the query inside `gather_schema_stats_oracle.sql` to run after all loading threads only on Oracle config for `chbenchmark` and `tpch`, due to massive performance issue. I noticed that I can override `LoaderThread.afterLoad()` and use `benchmarkModule.workConf.databaseType` to determine when to execute the query but ends up doing this. I can move the changes to `LoaderThread` if you want.

docker/oracle/reset.sql
ngbanguyen: I didn't think five lines of SQL is worth a checkout ([discussion](https://github.com/orgs/community/discussions/26359)). Maybe I can include it in the tar package?

src/main/java/com/oltpbenchmark/benchmarks/resourcestresser/procedures/CPU1.java
ngbanguyen: It should be for all DB type (for example [spanner](https://cloud.google.com/spanner/docs/reference/standard-sql/query-syntax#from_clause))

Oracle doesn't work if AS is kept, and it's not possible to use dialect file for these particular statements.

.github/workflows/maven.yml
```diff
@@ -10,6 +10,8 @@
 # - One job to build/test the docker images.
 # - One job to publish the docker image.
 
+# Copyright (c) 2023, Oracle and/or its affiliates.
+
```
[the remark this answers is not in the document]
ngbanguyen: Moved copyright header

config/oracle/sample_auctionmark_config.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0"?>
+<!--  Copyright (c) 2023, Oracle and/or its affiliates. -->
```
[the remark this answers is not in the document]
ngbanguyen: Moved copyright header

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/procedures/LoadConfig.java
ngbanguyen: Oracle DB DDL contains some CLOB fields (for LoadConfig procedures). These CLOB needs to be converted to String while the connection is alive.

This CLOB conversion for Oracle needs to be done here, otherwise the conversion will be attempted by `SQLUtil.getString(Object)` after the connection closes, which will result in `java.sql.SQLRecoverableException: Closed Connection`.

src/main/java/com/oltpbenchmark/util/SQLUtil.java
ngbanguyen: Oracle support needed TIMESTAMP fields which corresponds to OJDBC-specific type `oracle.sql.TIMESTAMP` (not available in JDBC). 

I need to convert OJDBC timestamp to JDBC `java.sql.Timestamp` (with [timestampValue](https://docs.oracle.com/en/database/oracle/oracle-database/21/jajdb/oracle/sql/TIMESTAMP.html#timestampValue__) method) without having to import the type directly, hence the use of reflection with `Class.forName(String)` and `getDeclaredMethod(String)`. For Oracle profile the types will be initialized and used in `getTimestamp(Object)` and for other profiles the variables will be null and the conversion branch wouldn't be reachable.

src/main/resources/benchmarks/tpch/gather_schema_stats_oracle.sql
ngbanguyen: Consolidated to `src/main/resources/benchmarks/gather_schema_stats_oracle.sql` now.

.github/workflows/maven.yml
ngbanguyen: This image is publicly and freely available via https://github.com/gvenzl/oci-oracle-free. It’s also used in other open-source projects like this Micronaut example application (https://blogs.oracle.com/java/post/json-relational-duality-views-with-micronaut-framework) and by the  testcontainers framework (https://java.testcontainers.org/modules/databases/oraclexe/).  The Oracle official container registry can be found here: https://container-registry.oracle.com/.

src/main/java/com/oltpbenchmark/api/Procedure.java
```diff
@@ -36,7 +37,15 @@ public abstract class Procedure {
...
-    private DatabaseType dbType;
+    /**
+     * LoadConfig procedure in auctionmark and seats benchmark contains CLOB on Oracle config
+     *
+     * For other databases, calling <code>SQLUtil.getString(Object)</code> after connection close is fine, but for
+     *      Oracle config, CLOBS needs to be resolved into String when the connection is open. Making this field
+     *      protected allows LoadConfig procedure to determine current database type, and decide if it should perform
+     *      the conversion.
+     */
+    protected DatabaseType dbType;
```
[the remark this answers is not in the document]
ngbanguyen: Added getter `getDbType()`

src/main/java/com/oltpbenchmark/benchmarks/otmetrics/procedures/GetSessionRange.java
ngbanguyen: Leaving this would break Oracle run for `otmetrics` because too many open connection due to not calling `stmt.close()` . Do I still remove this change and remove `otmetrics` from the GitHub Actions Oracle run?

src/main/resources/benchmarks/chbenchmark/gather_schema_stats_oracle.sql
ngbanguyen: For Oracle DB container spin up from Docker, some benchmarks requires gathering optimizer statistics in order to improve performance (especially tpch/chbenchmark with multiple 6+ way joins), otherwise time executed for each query would be too long.

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
ngbanguyen: For Oracle DB container spin up from Docker, some benchmarks requires gathering optimizer statistics in order to improve performance (especially tpch/chbenchmark with multiple 6+ way joins), otherwise time executed for each query would be too long. The optimizer query was invoked from this function.

Should I add test for this feature here, a separate PR, or just remove tpch + chbenchmark from GitHub Actions for Oracle?

ngbanguyen:
@apavlo @bpkroth I have added comments for some outstanding issues in the PR and in the code as well. Let me know if those are sufficient.

apavlo:
@ngbanguyen Thanks! I will look at this later in the week!

src/main/java/com/oltpbenchmark/benchmarks/resourcestresser/procedures/CPU1.java
bpkroth: We could certainly add switch statements on DBType if necessary.  Let's check the CI results, but as long as this works across all currently supported DBs I'm fine with the change as is.

src/main/java/com/oltpbenchmark/util/SQLUtil.java
```diff
@@ -109,20 +111,58 @@ public static Double getDouble(Object obj) {
...
+            // Then java.sql.Timestamp oracle.sql.TIMESTAMP.timestampValue() can be loaded
+            timestampValueMethod = oracleTimestamp.getDeclaredMethod("timestampValue");
+        } catch (ClassNotFoundException | NoSuchMethodException e) {
+            oracleTimestamp = null;
+            timestampValueMethod = null;
+        }
+        // If loading is successful then both variables won't be null.
+        TIMESTAMP_VALUE_METHOD = timestampValueMethod;
+        ORACLE_TIMESTAMP = oracleTimestamp;
+    }
```
bpkroth, line 165:
```suggestion
    }
    
```
nit: whitespace

CONTRIBUTORS.md
```diff
@@ -19,4 +19,5 @@
 We also acknowledge contributions from the following collaborators:
 
 + [Florian Funke](http://www3.in.tum.de/~funkef/)
-+ [Michael Seibold](http://www3.in.tum.de/~seibold/)
\ No newline at end of file
++ [Michael Seibold](http://www3.in.tum.de/~seibold/)
++ [Oracle](https://github.com/oracle) (Copyright (c) 2023, Oracle and/or its affiliates.)
```
bpkroth, line 23: Later we should consider to automate this and/or add additional authors from the history.

bpkroth: #390

docker/oracle/reset.sql
bpkroth: It strikes me that that script would be useful in generated docker images for benchbase too, so yes, I think we should include it in the package.  The provided sample configs in the image won't work without the file for instance.
Here's what I suggest:
- Move the file to `config/oracle/scripts/`
  Then, it should be included in the `tgz` by default.
- Update the various other callsites to reference that location instead.

src/main/java/com/oltpbenchmark/benchmarks/otmetrics/procedures/GetSessionRange.java
ngbanguyen: #396

src/main/java/com/oltpbenchmark/DBWorkload.java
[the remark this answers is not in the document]
ngbanguyen: #397

ngbanguyen:
I have added my explanation to unresolved conversation, and put it in the code as well. Are there anything else I need to do before resolving them?

Also, split some changes into PR #396 and #397. Please review @apavlo @bpkroth

bpkroth:
@ngbanguyen the two PRs that were split were merged.  Can you please update this branch from main to pull in the changes and resolve the conflicts so we can have another look at this PR?

Thanks!

src/main/java/com/oltpbenchmark/DBWorkload.java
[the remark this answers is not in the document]
ngbanguyen: Merged

.github/workflows/maven.yml
bpkroth: I'm not huge Oracle expert.
Is there a particular reason not to use the `container-registry.oracle.com/database/free:latest` image?
My intuition is that that would be more representative of a real Oracle DB for use in testing at least and is the officially supported upstream version.
What advantages or maintenance does the one provided by gvenzl offer?

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
```diff
@@ -15,7 +15,6 @@
  *
  */
 
-
```
bpkroth, line 18: nit: remove whitespace only changes

src/main/java/com/oltpbenchmark/api/Procedure.java
```diff
@@ -37,6 +37,18 @@ public abstract class Procedure {
 
     private final String procName;
     private DatabaseType dbType;
+    /**
```
bpkroth, line 40:
```suggestion

    /**
```

src/main/java/com/oltpbenchmark/api/Procedure.java
```diff
@@ -37,6 +37,18 @@ public abstract class Procedure {
...
+     * LoadConfig procedure in auctionmark and seats benchmark contains CLOB on Oracle config
+     *
+     * For other databases, calling <code>SQLUtil.getString(Object)</code> after connection close is fine, but for
+     *      Oracle config, CLOBS needs to be resolved into String when the connection is open. Making this field
+     *      protected allows LoadConfig procedure to determine current database type, and decide if it should perform
+     *      the conversion.
```
bpkroth, line 46:
```suggestion
     * the conversion.
```

nit: extra whitespace on these comment lines

bpkroth: But also, we probably don't need that comment here at all.  Only in the places where it's actually relevant (auctionmark and seats load procedures).

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkLoader.java
```diff
@@ -40,6 +40,8 @@
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
+import static com.oltpbenchmark.types.DatabaseType.ORACLE;
```
bpkroth, line 43:
```suggestion
import com.oltpbenchmark.types.DatabaseType;
```

and reference `DatabaseType.ORACLE` below would be more consistent with other parts of the code.

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkLoader.java
```diff
@@ -439,7 +441,9 @@ protected int populateRandomColumns(Object[] row) {
             // STRINGS
             for (Column catalog_col : this.random_str_cols) {
                 int size = catalog_col.getSize();
-                row[catalog_col.getIndex()] = profile.rng.astring(profile.rng.nextInt(size - 1), size);
+                // This (0) can generate an empty string which is treated as NULL in Oracle DB
+                int start = getDatabaseType() == ORACLE ? 1 : 0;
```
bpkroth, line 445:
```suggestion
                int start = getDatabaseType() == DatabaseType.ORACLE ? 1 : 0;
```

bpkroth:
Left some cleanup nits and one remaining comment about the docker image selection, but overall this is looking pretty good to me.  Thanks for your continued work on it!

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
ngbanguyen: #397

.github/workflows/maven.yml
ngbanguyen: The gvenzl image is slimmer (`gvenzl/oracle-xe:21.3.0-slim-faststart` at 4.63GB vs `container-registry.oracle.com/database/free:23.2.0.0` at 10.60GB through `docker inspect`) and starts slightly faster than official Oracle DB.

It also offers better maintenance. gvenzl image offers extra options `APP_USER` and `APP_USER_PASSWORD` to create a user on start up (similar to MySQL `APP_USER` and `APP_PASSWORD`) which is not supported by official Oracle DB image. On official Oracle image, in order to create the user I'd need to mount `reset.sql` to startup instead of setup due to an [open issue](https://github.com/oracle/docker-images/issues/2644).

bpkroth (approved):
LGTM.  @apavlo, any further comments?

apavlo (approved):
LGTM

PR #382, opened 2023-10-25
Fix Port Binding Issue During Testing

neeetman:
This PR addresses the issue where tests would fail if the designated port was already in use. 
```
[INFO ] 2023-10-25 07:18:47,185 [main]  com.oltpbenchmark.benchmarks.resourcestresser.TestResourceStresserBenchmark setUp - starting HSQLDB server for test [testGetTransactionType] on port [9099]
[Server@191774d6]: [Thread[HSQLDB Server @191774d6,5,main]]: run()/openServerSocket(): 
java.net.BindException: Address already in use
```
Changes:

- Modified the `setUp` method in `TestResourceStresserBenchmark` class.
- Added logic to try different ports using `ServerSocket` to find an available port.
- ~~The system will attempt up to `MAX_TRIES` to find an open port before throwing an exception.~~
- The the system will make a thorough effort in finding an available port.

### DIFF
--- a/src/test/java/com/oltpbenchmark/api/AbstractTestCase.java
+++ b/src/test/java/com/oltpbenchmark/api/AbstractTestCase.java
@@ -99,18 +99,33 @@ public abstract class AbstractTestCase<T extends BenchmarkModule> {
         HsqlProperties props = new HsqlProperties();
         //props.setProperty("server.remote_open", true);
 
-        int port = portCounter.incrementAndGet();
-
-        LOG.info("starting HSQLDB server for test [{}] on port [{}]", name.getMethodName(), port);
-
         server = new Server();
         server.setProperties(props);
         server.setDatabasePath(0, "mem:benchbase;sql.syntax_mys=true");
         server.setDatabaseName(0, "benchbase");
         server.setAddress("localhost");
-        server.setPort(port);
         server.setSilent(true);
         server.setLogWriter(null);
+
+        int port = -1;
+        boolean portFound = false;
+
+        for (int i = 0; i < MAX_TRIES; i++) {
+            port = portCounter.incrementAndGet();
+            try (ServerSocket testSocket = new ServerSocket(port)) {
+                portFound = true;
+                break;
+            } catch (BindException e) {
+                continue;
+            }   
+        }
+
+        if (!portFound) {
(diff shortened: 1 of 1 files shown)


apavlo:
@neeetman What problem is this trying to solve in testing? If the port is in use, then the test environment is wrong not the code.

neeetman:
> @neeetman What problem is this trying to solve in testing? If the port is in use, then the test environment is wrong not the code.

Each test requires applying for a new port. However, sometimes other services running on the server might already be occupying that port. It might be better to consider skipping the occupied ports instead of directly terminating the test with an error.

src/test/java/com/oltpbenchmark/api/AbstractTestCase.java
```diff
@@ -110,7 +113,7 @@ public final void setUp() throws Exception {
         server.setAddress("localhost");
         server.setPort(port);
         server.setSilent(true);
-        server.setLogWriter(null);
+        server.setLogWriter(null);
```
bpkroth, line 116:
```suggestion
        server.setLogWriter(null);
```

src/test/java/com/oltpbenchmark/api/AbstractTestCase.java
```diff
@@ -163,6 +166,22 @@ public final void setUp() throws Exception {
...
+            int port = portCounter.incrementAndGet();
+
+            if (port > MAX_PORT_NUMBER) {
+                throw new IOException("No available port found up to " + MAX_PORT_NUMBER);
+            }
+
+            try (ServerSocket testSocket = new ServerSocket(port)) {
+                return port;
+            } catch (BindException e) {
+                // This port is already in use. Continue to next port.
```
bpkroth, line 180: Please log a warning here before continuing.

bpkroth (approved):
LGTM.  Minor tweak suggested.

bpkroth:
@neeetman, just noticed you closed this before merging.  Any particular reason?

bpkroth:
> > @neeetman, just noticed you closed this before merging. Any particular reason?
> 
> Sorry, I previously noticed that 'Some checks were not successful,' so I closed it.

Ah, for future, just ping back about that.  Happy to help look into those.

That one looks spurious.  Seems to have failed in the nightly run a few times as well, so probably needs separate investigation.  I'll file an issue on that one and follow up separately.

Issue #365, opened 2023-09-26
Add CI checks for auto generated code.


bpkroth:
Per discussions in #301, there are a few places where we auto-generate code.
It would be useful to have a check in the CI pipelines that these files have not been edited by hand and are up to date.

> Can we automate this too?  e.g. run it during the CI pipelines and make sure that no file changes are generated (e.g. with a `git diff --porcelain` check and abort the checkin since local changes are out of sync.

_Originally posted by @bpkroth in https://github.com/cmu-db/benchbase/pull/301#discussion_r1197975781_

PR #385, opened 2023-10-31
Feat: Support for Distributions in Templated Workloads

ETHenzlere:
PR adds support for various Numerical Distributions in templated benchmarks

- Uniform
- Binomial
- Zipfian
- Scrambled

The distributions work not only for integers, but also different types such as timestamps, long, float
A table of supported combinations can be found in the templated benchmarks readme.

| Type | uniform | binomial | zipfian | scrambled (zipfian) |
|---|:---:|:---:|:---:|:---:|
| INTEGER | X | X | X | X |
| FLOAT / REAL | X | X|- | - |
| BIGINT | X | X | X | X |
| VARCHAR / STRING | X | -| -| -|
| TIMESTAMP | X | X |X |X |
| DATE | X | X| X| X|
| TIME | X |X | X| X|

Usage example:

```xml
<templates>
    <template name="MyTemplate">
        <query><![CDATA[SELECT * FROM MyTable WHERE id = ?]]></query>
        <types>
            <type>INTEGER</type>
        </types>
        <values>
            <value dist="uniform" min="0" max="1000" seed="1"/>
        </values>
         <values>
            <value dist="zipf" min="1" max="1000" seed="2"/>
        </values>
    </template>
    <!-- ... -->
<templates>
```
This PR is not a breaking change - One can still use a static value in the templated queries.
`<value>10</value>`

In the future, I could see a breaking change that adds the datatype directly to the values so the TemplatedValue can do the type handling directly. `<value dist="uniform" type="integer">`. This would make type checking easier and remove the need to store the original min/max values as strings for all datatypes.


data/templated/example.xml
```diff
@@ -20,10 +20,10 @@
    <template name="GetCust">
       <query><![CDATA[SELECT C_DISCOUNT, C_LAST, C_CREDIT FROM customer WHERE C_W_ID = ?]]></query>
       <types>
-         <type>INTEGER</type>
+         <type>DISTRIBUTION</type>
```
anjagruenheid, line 23: Don't replace, add another template?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,33 +56,98 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+                    case "timestamp":
+                        Timestamp timestamp = new Timestamp(System.currentTimeMillis());
+                        stmt.setTimestamp(i + 1, timestamp);
+                        break;
+                    case "date":
+                        Date date = new Date(System.currentTimeMillis());
+                        stmt.setDate(i + 1, date);
+                        break;
+                    case "time":
+                        Time time = new Time(System.currentTimeMillis());
```
anjagruenheid, line 128: Inline below

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,33 +58,96 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+                        break;
+                    case "date":
+                        stmt.setDate(i + 1, new Date(System.currentTimeMillis()));
+                        break;
+                    case "time":
+                        stmt.setTime(i + 1, new Time(System.currentTimeMillis()));
+                        break;
+                    default:
+                        throw new RuntimeException(
+                                "No suitable distribution found. Currently supported are 'zipf' | 'scrambled' | 'normal' | 'uniform' | 'string' | 'datetime' | 'timestamp' | 'date' | 'time' ");
```
bpkroth, line 134: 👍

data/templated/example.xml
```diff
@@ -61,4 +61,16 @@
          <value>11</value>
       </values>
    </template>
+   <template name="GetItemRandom">
+      <query><![CDATA[SELECT COUNT(*) FROM item WHERE i_im_id = ?]]></query>
+      <types>
+         <type>RANDOM</type>
+      </types>
+      <values>
+         <value>binomial,0,9999</value>
```
bpkroth, line 70: Would be good to have a test case for each of your supported types.
That would also help document a little better how this should work.
Another nice documentation item could be to expand on the `README.md` in the templated benchmarks source directory.

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,33 +58,96 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+                        break;
+                    case "scrambled":
+                        ScrambledZipfianGenerator scramZipf = (ScrambledZipfianGenerator) randomGenerators.get(index);
+                        stmt.setInt(i + 1, scramZipf.nextInt());
+                        j += 2;
+                        break;
+                    case "string":
+                        maxI = Integer.parseInt(params.get(index + 1).toString());
+                        Random rand = (Random) randomGenerators.get(index);
+                        String randText = TextGenerator.randomStr(rand, maxI);
```
bpkroth, line 118: This would only support uniform distribution for strings, correct?
Wouldn't it be more useful to specify 
a) the type of the param (string, int, float), and
b) the distribution (and parameters) of that type separately?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,33 +58,96 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+                        break;
+                    case "datetime":
+                    case "timestamp":
+                        stmt.setTimestamp(i + 1, new Timestamp(System.currentTimeMillis()));
+                        break;
+                    case "date":
+                        stmt.setDate(i + 1, new Date(System.currentTimeMillis()));
+                        break;
+                    case "time":
+                        stmt.setTime(i + 1, new Time(System.currentTimeMillis()));
```
bpkroth, line 130: These time types aren't actually `RANDOM`, right?
Again, it'd be useful to be able to specify random time ranges for many benchmarks.

src/main/java/com/oltpbenchmark/benchmarks/templated/util/GenericQueryOperation.java
```diff
@@ -18,24 +18,68 @@
...
+    private HashMap<Integer, Object> mapRandom() {
+        HashMap<Integer, Object> randomGenerators = new HashMap<Integer, Object>();
+        Integer min, max;
+        for (int i = 0; i < params.size(); i++) {
+            String paramString = params.get(i).toString();
+            switch (paramString) {
+                case "zipf":
+                    min = Integer.parseInt(params.get(i + 1).toString());
+                    max = Integer.parseInt(params.get(i + 2).toString());
+                    ZipfianGenerator zipf = new ZipfianGenerator(new Random(), min, max);
```
bpkroth, line 56: What about specifying seeds to make these reproducible?

data/templated/example.xml
```diff
@@ -61,4 +61,16 @@
...
+      </values>
+      <values>
+         <value>uniform,0,9999</value>
```
bpkroth, line 73: Per commentary below, I think the representation here should be expanded to something more descriptive than a bar CSV.

For instance:
```xml
<types>
  <!-- maintain the actual parameter type here -->
  <type>integer</type>
</types>
<values>
  <!-- instead of specifying a <value> instance here, specify a new xml element - <value_distribution> -->
  <value_distribution type="binomial" min="0" max="9999" seed="42" />
  <!-- 
    add any other params needed for that distribution type in the properties of that value_distribution
    seed could be omitted and pulled from the global seed in that case, but this would allow individually selecting the starting seed for each parameter
  -->
</values>
```

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,36 +58,129 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+
+            ComplexValue param = params.get(i);
+            boolean hasDist = param.getDist().length() > 0;
+            boolean hasValue = param.getValue().length() > 0;
+
+            if ((!hasDist && !hasValue) || paramsTypes[i].equalsIgnoreCase("NULL")) {
                 stmt.setNull(i + 1, Types.NULL);
+            
+            } else if (hasDist) {
+                String distribution = param.getDist();
```
ETHenzlere, line 85: @bpkroth I made some changes to the architecture to better support the distributions. Would this be something you feel like you could work with?

The new system would still take the <value> element but now allows to choose a distribution like this:
<value dist="uniform" min="0" max="10" seed="0" />
While still working with a single value: <value>10</value> and also NULL type: <value />
(So no breaking change)

So far I have only implemented the uniform distribution because I do not want to change everything again if there is feedback to the Architecture :)

data/templated/example.xml
ETHenzlere: Will do as soon as I publish the PR :+1:

anjagruenheid: I think the documentation should be updated as part of the PR to be consistent.

src/main/java/com/oltpbenchmark/benchmarks/templated/util/ComplexValue.java
```diff
@@ -0,0 +1,82 @@
...
+        this.minS = min;
+        this.maxS = max;
+    }
+
+    private Object createGen(String distribution, Long min, Long max, Long seed) {
+
+        if(distribution == null || distribution.length() < 1) return null;
+
+        switch (distribution) {
+            case "uniform":
```
anjagruenheid, line 36: I think I'd prefer these as enums to ensure that they're consistent. What do you think?

src/main/java/com/oltpbenchmark/benchmarks/templated/util/ComplexValue.java
```diff
@@ -0,0 +1,82 @@
...
+        this.min = min == null ? 0 : Long.parseLong(min);
+        this.max = max == null ? 1 : Long.parseLong(max);
+        this.seed = seed == null ? 0 : Long.parseLong(seed);
+        this.value = value;
+        this.randGen = createGen(dist,this.min,this.max,this.seed);
+        this.minS = min;
+        this.maxS = max;
+    }
+
+    private Object createGen(String distribution, Long min, Long max, Long seed) {
```
anjagruenheid, line 31: Nit: createGenerator?

src/main/java/com/oltpbenchmark/benchmarks/templated/util/ComplexValue.java
```diff
@@ -0,0 +1,82 @@
...
+    public final Object randGen;
+
+
+    public ComplexValue(String dist, String min, String max, String seed, String value) {
+        this.dist = dist == null ? "" : dist;
+        this.min = min == null ? 0 : Long.parseLong(min);
+        this.max = max == null ? 1 : Long.parseLong(max);
+        this.seed = seed == null ? 0 : Long.parseLong(seed);
+        this.value = value;
+        this.randGen = createGen(dist,this.min,this.max,this.seed);
```
anjagruenheid, line 26: Nit: why did you name it randGen? generator is more descriptive.

src/main/java/com/oltpbenchmark/benchmarks/templated/util/ComplexValue.java
```diff
@@ -0,0 +1,82 @@
+package com.oltpbenchmark.benchmarks.templated.util;
+
+import java.util.Random;
+
+import com.oltpbenchmark.distributions.ScrambledZipfianGenerator;
+import com.oltpbenchmark.distributions.ZipfianGenerator;
+
+public class ComplexValue {
```
anjagruenheid, line 8: class documentation?

anjagruenheid: and why is it called ComplexValue?

src/main/java/com/oltpbenchmark/api/templates/TemplatesType.java
```diff
@@ -15,60 +15,77 @@
...
+     * {@link TemplateType }
+     * 
      * 
-     * Objects of the following type(s) are allowed in the list {@link TemplateType }
      */
-    public List<TemplateType> getTemplateList() {
-        if (this.template == null) {
-            this.template = new ArrayList<>();
+    public List<TemplateType> getTemplate() {
+        if (template == null) {
```
anjagruenheid, line 85: this.template?

src/main/java/com/oltpbenchmark/api/templates/TypesType.java
```diff
@@ -16,55 +16,74 @@
...
+     * Objects of the following type(s) are allowed in the list
+     * {@link String }
+     * 
+     * 
      */
-    public List<String> getTypeList() {
-        if (this.type == null) {
-            this.type = new ArrayList<>();
+    public List<String> getType() {
+        if (type == null) {
```
anjagruenheid, line 83: s.a.

src/main/java/com/oltpbenchmark/api/templates/TypesType.java
```diff
@@ -16,55 +16,74 @@
...
+            type = new ArrayList<String>();
```
anjagruenheid, line 84: Nit: no need to specify that it's a String :)

src/main/resources/templates.xsd
```diff
@@ -46,8 +46,22 @@
     <!-- <values> -->
     <xsd:complexType name="valuesType">
         <xsd:sequence>
-            <xsd:element name="value" type="xsd:string" minOccurs="1" maxOccurs="unbounded"/>
+            <xsd:element name="value" type="valueType" minOccurs="1" maxOccurs="unbounded"/>
         </xsd:sequence>
```
anjagruenheid, line 50: Nit: indentation looks a bit weird here

data/templated/example.xml
```diff
@@ -61,4 +61,16 @@
          <value>11</value>
       </values>
    </template>
+   <template name="GetItemRandom">
+      <query><![CDATA[SELECT COUNT(*) FROM item WHERE i_im_id = ?]]></query>
+      <types>
+         <type>INTEGER</type>
+      </types>
+      <values>
+         <value dist="uniform" min="0" max="9999" seed="12"></value>
```
anjagruenheid, line 70: Please add tests for all possible distributions.

config/postgres/sample_templated_config.xml
bpkroth, line 1: Similar comment to the one given in #386: please expand this sample_templated_config.xml to all of the supported DB types so that we can check this change over all DB types in the CI pipelines.
Thanks!

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
```diff
@@ -54,3 +55,47 @@ java -jar benchbase.jar -b templated -c config/sqlserver/sample_templated_config
...
+The distributions are dependent on the type of the value. Currently, the following type-distributions pairs are supported:
+| Type | uniform | binomial | zipfian | scrambled zipfian |
+|---|:---:|:---:|:---:|:---:|
+| INTEGER | X | X | X | X |
+| FLOAT / REAL | X | X|- | - |
+| BIGINT | X | X | X | X |
+| VARCHAR / STRING | X | -| -| -|
+| TIMESTAMP | X | X |X |X |
+| DATE | X | X| X| X|
+| TIME | X |X | X| X|
```
bpkroth, line 92: nit about the whitespace - aligning these column markers will make it easier to edit the markdown in the future

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
```diff
@@ -54,3 +55,47 @@ java -jar benchbase.jar -b templated -c config/sqlserver/sample_templated_config
...
+
+The following properties can be set on the value:
+
+-   dist _The distribution of the values_
```
bpkroth, line 96:
```suggestion
-   `dist`: The distribution of the values.
```

style nit (same for the rest of them)

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -170,38 +181,44 @@ private CustomClassLoader loadQueryTemplates(String file) {
             JAXBElement<TemplatesType> result = unmarshaller.unmarshal(streamSource, TemplatesType.class);
             TemplatesType templates = result.getValue();
 
-            for (TemplateType template : templates.getTemplateList()) {
+            for (TemplateType template : templates.getTemplate()) {
```
bpkroth, line 184: Why not `getTemplateList()` anymore?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -170,38 +181,44 @@ private CustomClassLoader loadQueryTemplates(String file) {
...
                 ImmutableParsedQueryTemplate.Builder b = ImmutableParsedQueryTemplate.builder();
                 b.name(template.getName());
                 b.query(template.getQuery());
-                b.paramsTypes(template.getTypes().getTypeList());
+                b.paramsTypes(template.getTypes().getType());
```
bpkroth, line 188: Similar: why not `getTypeList()` anymore?

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
ETHenzlere: All of these files were automatically generated but I will make sure to change it :)

ETHenzlere:
I am waiting for my other PR, #386  to be merged. This will allow me to make less changes to the query templates in order to test all different combinations of types and distributions.

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
bpkroth: I was just looking for an explanation actually.  I don't think you should go editing auto-generated code after that fact.    That leads to a maintenance nightmare.  Please revert it to whatever the standard auto generated code was for now.

Even better would be that the auto-generated code gets auto-generated during build time without extra steps so it can automatically rev with dependencies and we know when there are issues, but I think that's a separate issue (#365).

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedWorker.java
bpkroth, line 1: nit: generally good not to submit whitespace only changes to files in a PR.

I agree they're nice to get done where relevant, but it can really muddy up reviews and histories.

I wouldn't be sad at all to see some sort of automatic linter/formatter added to the CI pipelines and pre-commit-hooks, but introducing that would likely change many files and should probably be kept to its own PR.

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,37 +59,245 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+            } else if (hasDist) {
+                ValueGenerator distribution = param.getDistribution();
+                String paramType = paramsTypes[i].toLowerCase();
+                switch (paramType) {
+                    case "integer":
+                        int min = param.getMin().intValue();
+                        int max = param.getMax().intValue();
+
+                        switch (distribution) {
+
```
bpkroth, line 94:
```suggestion
```

nit: extra unbalanced whitespace

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,37 +59,245 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
         for (int i = 0; i < paramsTypes.length; i++) {
-            if (paramsTypes[i].equalsIgnoreCase("NULL")) {
+
+            TemplatedValue param = params.get(i);
+            boolean hasDist = param.getDistribution() != null;
+            boolean hasValue = param.getValue().length() > 0;
+
+            if ((!hasDist && !hasValue) || paramsTypes[i].equalsIgnoreCase("NULL")) {
                 stmt.setNull(i + 1, Types.NULL);
+
```
bpkroth, line 84:
```suggestion
```

nit: unbalanced whitespace

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,37 +59,245 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
 
-    public PreparedStatement getStatement(Connection conn, List<Object> params) throws SQLException {
+    public PreparedStatement getStatement(Connection conn, List<TemplatedValue> params) throws SQLException {
         QueryTemplateInfo queryTemplateInfo = this.getQueryTemplateInfo();
 
         PreparedStatement stmt = this.getPreparedStatement(conn, queryTemplateInfo.getQuery());
         String[] paramsTypes = queryTemplateInfo.getParamsTypes();
         for (int i = 0; i < paramsTypes.length; i++) {
-            if (paramsTypes[i].equalsIgnoreCase("NULL")) {
+
```
bpkroth, line 77:
```suggestion
```

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * 
+     * @param dist  The desired value distribution
+     * @param min   Minimum value. Default is 0
+     * @param max   Maximum value. Default is 1
+     * @param seed  The seed for the random generator. Default is 0
+     * @param value Value that is used if no distribution is given
+     */
+    public TemplatedValue(String dist, String min, String max, String seed, String value) {
```
bpkroth, line 32:
```suggestion
    public TemplatedValue(String distribution, String min, String max, String seed, String value) {
```

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,112 @@
...
+        this.min = min == null ? 0 : Long.parseLong(min);
+        this.max = max == null ? 1 : Long.parseLong(max);
+        this.seed = seed == null ? 0 : Long.parseLong(seed);
+        this.value = value;
```
bpkroth, line 36: Is `value` valid when `distribution` is also provided?

bpkroth: General point is you probably need some validation of the values being passed in (e.g., all expected ones are present or not, correct relative ranges, etc.)

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,112 @@
...
+
+    public final Object generatorObject;
+
+    /**
+     * 
+     * @param dist  The desired value distribution
+     * @param min   Minimum value. Default is 0
+     * @param max   Maximum value. Default is 1
+     * @param seed  The seed for the random generator. Default is 0
+     * @param value Value that is used if no distribution is given
```
bpkroth, line 30: If `value` is the more common expectation, I might suggest to provide that as the first element.

Another suggestion would be to create to constructors that call into this (private) one:
- One with distribution related content.
- The other with just a value.

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,112 @@
...
+ * generators for value distributions
+ */
+public class TemplatedValue {
+    ValueGenerator distribution;
+    Long min;
+    Long max;
+    Long seed;
+    String value;
+    String minS;
+    String maxS;
```
bpkroth, line 20: Why store `minS` and `maxS` from just the `min` and `max` `long` values?

Can't you just `toString()` those as necessary?

Or was this intended to support `min` `max` string for String type distributions?
In that case, you'd need to always pass the String value in instead correct?
Maybe this is a case for a generic class that abstracts the internal type stored (e.g., `TemplatedValue<Long>` vs. `TemplatedValue<String>`

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,112 @@
...
+    public Long getMax() {
+        return this.max;
+    }
+
+    public Long getSeed() {
+        return this.seed;
+    }
+
+    public String getValue() {
+        return this.value;
```
bpkroth, line 98:
```suggestion
        assert this.distribution == null;
        return this.value;
```

At least in debug mode.
Similar for the others above.

src/main/java/com/oltpbenchmark/benchmarks/templated/util/GenericQueryOperation.java
```diff
@@ -27,15 +26,14 @@
...
 
-    public final List<Object> params;
+    public final List<TemplatedValue> params;
 
-
-    public GenericQueryOperation(Object[] params) {
+    public GenericQueryOperation(TemplatedValue[] params) {
         super();
-        this.params = Collections.unmodifiableList(Arrays.asList(params));
+        this.params = Arrays.asList(params);
```
bpkroth, line 33: Why did the unmodifiable part get dropped?

src/main/java/com/oltpbenchmark/benchmarks/templated/util/GenericQueryOperation.java
```diff
@@ -27,15 +26,14 @@
...
     }
 
-    public List<Object> getParams() {
+    public List<TemplatedValue> getParams() {
```
bpkroth, line 36: 👍

src/main/java/com/oltpbenchmark/benchmarks/templated/TemplatedBenchmark.java
```diff
@@ -170,38 +181,44 @@ private CustomClassLoader loadQueryTemplates(String file) {
...
-                            getParamsString(qt.getParamsValues()));
+                        GenericQuery.class.getPackageName(),
+                        TemplatedValue.class.getCanonicalName(),
+                        qt.getName(),
+                        GenericQuery.class.getCanonicalName(),
+                        QueryTemplateInfo.class.getCanonicalName(),
+                        SQLStmt.class.getCanonicalName(),
+                        StringEscapeUtils.escapeJava(qt.getQuery()),
+                        getParamsString(qt.getParamsTypes()),
+                        buildTemplatedValueString(qt.getParamsValues()));
```
bpkroth, line 221: indentation change here makes it a little difficult to see the actual change

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -48,37 +59,245 @@ public void run(Connection conn, List<Object> params) throws SQLException {
...
+                        }
+                        break;
+                    case "time":
+                        Long timeMin = param.getMin();
+                        Long timeMax = param.getMax();
+                        switch (distribution) {
+                            case UNIFORM:
+                                Random uniformRand = (Random) param.getGenerator();
+                                stmt.setTime(i + 1, new Time(uniformRand.nextLong(timeMin, timeMax)));
+                                break;
```
bpkroth, line 258: There is a lot of repetition in many of these blocks.

For starters, in each block, the `stmt.setSomeType(i + 1, generatedValue)` portion is repeated.  It could just go at the end of the internal switch block.

But once you do that, I think you'll find that many of the internal distribution switch statements are repeated across many of the types.
For instance, most of them just look up and cast the appropriate distribution type, then return a value within the specified range using a `long`.  I think `ints` could be handled using the `long` code and just throwing an exception either on construction time when the range is first given (best option since the error code only happens once and early then) or at runtime when the `long` value generated is not castable into the `int` range.

That leaves floats and strings as the only exceptions.

Anyways, I think all of that `genNextValue` could probably be abstracted into the `TemplatedValue` class and called at the top of this block here so that mostly the only thing this switch statement is casting the generated value to the appropriate sql type when setting the parameter.

bpkroth:
> I am waiting for my other PR, #386 to be merged. This will allow me to make less changes to the query templates in order to test all different combinations of types and distributions.

Makes sense.  I think there's only one small issue remaining there and then we can merge that one.  For now I left some more comments here as well.

bpkroth:
> > I am waiting for my other PR, #386 to be merged. This will allow me to make less changes to the query templates in order to test all different combinations of types and distributions.
> 
> Makes sense. I think there's only one small issue remaining there and then we can merge that one. For now I left some more comments here as well.

FWIW, that one got merged, though I later discovered a bug with it: #428 

Also, #416 added automatic style changes, so you'll need to rebase your changes on an updated main first.

ETHenzlere:
I will adapt the PR with the fix from the latest issue and then it should be ready to go out of draft mode

bpkroth:
> I will adapt the PR with the fix from the latest issue and then it should be ready to go out of draft mode

Sounds good.  Do please try and merge in the latest main and fix the conflicts.  Thanks!

data/templated/example.xml
```diff
@@ -97,9 +97,9 @@
          <value>1</value>
          <value>1</value>
          <value>1</value>
-         <value>2022-10-10 11:30:30</value>
-         <value>1.0</value>
-         <value>Test</value>
+         <value dist="uniform" min="1674477220000" max="1706013220000"/>
```
bpkroth, line 100: Why no test for distributions of timestamp types?  That was supported, right?

bpkroth: Rather, what's the unit on this min/max?  Can we not parse actual timestamps?

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
```diff
@@ -54,3 +55,46 @@ java -jar benchbase.jar -b templated -c config/sqlserver/sample_templated_config
...
+| DATE             |    X    |    X     |    X    |         X         |
+| TIME             |    X    |    X     |    X    |         X         |
+The following properties can be set on the value:
+
+-   `dist`: The distribution of the values
+-   `min`: The minimum value the generator can produce. Defaults to 0
+-   `max`: The maximum value the generator can produce. Defaults to 1
+-   `seed`: A seed for the generator to ensure consistency
+
+For Timestamps, Dates and Times, the min and max values must be converted to a long (milliseconds)
```
bpkroth, line 100:
```suggestion
*NOTE:* For `Timestamp`, `Date`, and `Time` types, the min and max values must be converted to a long (milliseconds).
```

bpkroth: This can be a little cumbersome to read and do.  Is this number of milliseconds since unix epoch utc, for instance?

Ideally, we can handle the conversion for them using a datetime parser when instantiating.

If that's too difficult, can we please at least provide documentation examples of how to convert timestamps of the various formats using standard command line tools like `date`?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -76,32 +81,139 @@ public void run(Connection conn) throws SQLException {
...
+      boolean hasValue = param.getValue() != null;
+
+      if ((!hasDist && !hasValue) || paramsTypes[i].equalsIgnoreCase("NULL")) {
         stmt.setNull(i + 1, Types.NULL);
+
+      } else if (hasDist) {
+        ValueGenerator distribution = param.getDistribution();
+        String paramType = paramsTypes[i].toLowerCase();
+        switch (paramType) {
+          case "integer":
```
bpkroth, line 103: I think we also discussed changing these static strings into an enum.

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,178 @@
...
+      case "binomial":
+      case "normal":
+        this.distribution = ValueGenerator.BINOMIAL;
+        return new Random(seed);
+      case "zipf":
+      case "zipfian":
+        this.distribution = ValueGenerator.ZIPFIAN;
+        return new ZipfianGenerator(new Random(seed), min, max);
+      case "scrambled":
+      case "scramzipf":
```
bpkroth, line 90: Can't we use enums to map to these strings too?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -76,32 +81,139 @@ public void run(Connection conn) throws SQLException {
...
+              stmt.setDate(i + 1, new Date(generatedTimestamp));
+            } else {
+              stmt.setTime(i + 1, new Time(generatedTimestamp));
+            }
+            break;
+          default:
+            throw new RuntimeException(
+                "Support for distributions for the type: "
+                    + paramType
+                    + " is current not implemented");
```
bpkroth, line 200:
```suggestion
                    + " is not currently implemented");
```

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -76,32 +81,139 @@ public void run(Connection conn) throws SQLException {
...
+        }
+
       } else {
         try {
-          // TODO: add support for nullable other types
-          // For instance, can we provide a <value /> tag in the XML file to represent a
-          // NULL value?
-          // Or does it need a special marker like "$null" to signify a NULL value?
```
bpkroth, line 92: Do we have tests for NULL value handling now?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -76,32 +81,139 @@ public void run(Connection conn) throws SQLException {
...
     for (int i = 0; i < paramsTypes.length; i++) {
-      if (paramsTypes[i].equalsIgnoreCase("NULL")) {
+
+      TemplatedValue param = params.get(i);
+      boolean hasDist = param.getDistribution() != null;
+      boolean hasValue = param.getValue() != null;
+
+      if ((!hasDist && !hasValue) || paramsTypes[i].equalsIgnoreCase("NULL")) {
         stmt.setNull(i + 1, Types.NULL);
+
```
bpkroth, line 98: nit:

```suggestion
```

bpkroth:
Looking pretty good.  Just a few cosmetic polish things left:
- documentation or improved handling of timestamps as longs issue in the configs
- enums instead of static strings
- expanded test coverage for a few cases

data/templated/example.xml
ETHenzlere: I tried to find a clean and lightweight solution but could not come up with something satisfying. Since the templated value has no information about the type, it would require a nasty try-catch solution. Also, different time formats would have to be supported if one wants to do it right. **For now, I suggest to use UNIX time in ms as a common ground**.

In the future, I could also imagine to add a new distribution called "CurrentTime" or something like that, that always uses the current UTC and is only available for Timestamp/Date/Time types. This would make it easier for this specific use-case.

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
ETHenzlere: Added documentation :+1:

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
ETHenzlere: Yes, the system test how to handle multiple cases (Explicitly stating a value to be of type NULL or having an empty <value /> tag).

data/templated/example.xml
bpkroth: Not sure I follow this comment.
The type of the field is known from the `<types>` section, right?
A small try/catch during config load time (not query runtime) to convert them into timestamps seems pretty straight forward to me.

```java
int get_timestamp(String str) {
 int ts;
  try {
    ts = Integer.parseInt(param_value.min)
  }
  catch (NumberFormatException ex) {
    try {
      ts = DateTimeFormatter.parse(param_value.min).getTime()
    }
    catch (DateTimeParseException ex2) {
      // parse error - report to user that their config is malformed
      throw ex;
    }
  }
  return min_ts;
}

if (param_type == TIMESTAMP) {
  int min_ts = get_timestamp(param_value.min);
  int max_ts = get_timestamp(param_value.max);
}
```

No?

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  /**
+   * @param distribution The desired value distribution
+   * @param min Minimum value. Default is 0
+   * @param max Maximum value. Default is min + 1
+   * @param seed The seed for the random generator. Default is 0
+   * @param value Value that is used if no distribution is given
+   */
+  public TemplatedValue(String distribution, String min, String max, String seed) {
```
bpkroth, line 41: Re datetime parsing, couple options I can see here:
- Make different generic types for TemplateValue (probably extra work)
- Pass the "param_type" in with the constructor and use it do the extra min/max parsing logic, as well as some sanity check asserts for the various `getNext...` calls (e.g., get next string, for a non-string type is probably a mistake).

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -0,0 +1,184 @@
...
+    try {
+      this.min = Long.parseLong(min);
+    } catch (Exception e) {
+      this.min = 0L;
```
bpkroth, line 45: This should probably return an error, no?

data/templated/example.xml
ETHenzlere: Added type information and added DateTimeFormatter. The specific format for timestamps is (yyyy-MM-dd HH:mm:ss)

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
ETHenzlere: I added the param type to the TemplatedValue and changed the parse logic now

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
ETHenzlere: Used it as a fallback but now changed such that min and max are always required!

src/main/java/com/oltpbenchmark/benchmarks/templated/util/TemplatedValue.java
```diff
@@ -108,27 +170,35 @@ public ValueGenerator getDistribution() {
...
+  public String getValueType() {
+    return this.valueType.toString();
+  }
+
   public Object getGenerator() {
     return this.generatorObject;
   }
 
   public Long getNextLongBinomial() {
+    assert !this.valueType.equals(JDBCSupportedType.REAL)
```
bpkroth, line 198: Probably better to assert it's one of a known set of working values.

src/main/java/com/oltpbenchmark/benchmarks/templated/README.md
```diff
@@ -93,17 +93,17 @@ The distributions are dependent on the type of the value. Currently, the followi
...
--   `min`: The minimum value the generator can produce. Defaults to 0
--   `max`: The maximum value the generator can produce. Defaults to 1
--   `seed`: A seed for the generator to ensure consistency
+-   `min`: The minimum value the generator can produce.
+-   `max`: The maximum value the generator can produce.
+-   `seed`: A seed for the generator to ensure consistency. Default is 0
 
-*NOTE:* For `Timestamp`, `Date`, and `Time` types, the min and max values must be converted to UNIX time (milliseconds).
+*NOTE:* For `Timestamp`, `Date`, and `Time` types, the min and max values must be converted to UNIX time (milliseconds) OR given as strings in the following format: (yyyy-MM-dd HH:mm:ss).
 The value is interpreted as the milliseconds since January 1, 1970, 00:00:00 GMT/UTC.
```
bpkroth, line 101:
```suggestion
The int value is interpreted as the milliseconds since January 1, 1970, 00:00:00 GMT/UTC.
The string value is also interpreted as UTC timezone.
```

bpkroth: True?

data/templated/example.xml
```diff
@@ -106,9 +106,9 @@
          <value>1</value>
          <value>1</value>
          <value>1</value>
-         <value dist="uniform" min="1674477220000" max="1706013220000"/>
+         <value dist="uniform" min="1999-12-12 01:01:55" max="1706013220000"/>
```
bpkroth, line 109: FWIW, this example is a little weird.  Are mixed min/max string vs. int specification types allowed?  I suppose so, but might make more sense to provide two separate examples:
- one with both as ints
- one with both as strings

ETHenzlere: Switched this to string only. Example with int/long already exists

bpkroth (approved):
LGTM.  Thanks!

PR #386, opened 2023-10-31
Feat: Support for "Update" type queries in Templated Benchmarks

ETHenzlere:
Feature adds a check for UPDATE style queries and handles the execution in templated benchmarks

### DIFF
--- a/config/mariadb/sample_templated_config.xml
+++ b/config/mariadb/sample_templated_config.xml
@@ -18,7 +18,7 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
-        <!-- The workload -->
+    <!-- The workload -->
     <terminals>1</terminals>
     <works>
         <work>
--- a/config/mariadb/sample_templated_config.xml
+++ b/config/mariadb/sample_templated_config.xml
@@ -18,13 +18,13 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
-    <!-- The workload -->
+        <!-- The workload -->
     <terminals>1</terminals>
     <works>
         <work>
             <time>10</time>
             <rate>100</rate>
-            <weights>30,20,10,30,10</weights>
+            <weights>20,20,10,10,10,10,10,10</weights>
         </work>
     </works>
 
@@ -44,6 +44,15 @@
         </transactiontype>
         <transactiontype>
             <name>GetItemByPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>UpdateItemPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>DeleteItem</name>
+        </transactiontype>
+        <transactiontype>
+            <name>InsertItem</name>
         </transactiontype>
     </transactiontypes>
 </parameters>
--- a/config/postgres/sample_templated_config.xml
+++ b/config/postgres/sample_templated_config.xml
@@ -24,7 +24,7 @@
         <work>
             <time>10</time>
             <rate>100</rate>
-            <weights>30,20,10,25,10,5</weights>
+            <weights>30,20,10,20,10,10</weights>
         </work>
     </works>
 
--- a/config/postgres/sample_templated_config.xml
+++ b/config/postgres/sample_templated_config.xml
@@ -24,7 +24,7 @@
         <work>
             <time>10</time>
             <rate>100</rate>
-            <weights>30,20,10,30,10</weights>
+            <weights>30,20,10,25,10,5</weights>
         </work>
     </works>
 
@@ -44,6 +44,9 @@
         </transactiontype>
         <transactiontype>
             <name>GetItemByPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>UpdateItemPrice</name>
         </transactiontype>
     </transactiontypes>
 </parameters>
--- a/config/postgres/sample_templated_config.xml
+++ b/config/postgres/sample_templated_config.xml
@@ -24,7 +24,7 @@
         <work>
             <time>10</time>
             <rate>100</rate>
-            <weights>30,20,10,20,10,10</weights>
+            <weights>20,20,10,10,10,10,10,10</weights>
         </work>
     </works>
 
@@ -47,6 +47,12 @@
         </transactiontype>
         <transactiontype>
             <name>UpdateItemPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>DeleteItem</name>
+        </transactiontype>
+        <transactiontype>
+            <name>InsertItem</name>
         </transactiontype>
     </transactiontypes>
 </parameters>
--- a/config/sqlserver/sample_templated_config.xml
+++ b/config/sqlserver/sample_templated_config.xml
@@ -18,7 +18,7 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
-        <!-- The workload -->
+    <!-- The workload -->
     <terminals>1</terminals>
     <works>
         <work>
--- a/config/sqlserver/sample_templated_config.xml
+++ b/config/sqlserver/sample_templated_config.xml
@@ -18,13 +18,13 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
-    <!-- The workload -->
+        <!-- The workload -->
     <terminals>1</terminals>
     <works>
         <work>
             <time>10</time>
             <rate>100</rate>
-            <weights>30,20,10,30,10</weights>
+            <weights>20,20,10,10,10,10,10,10</weights>
         </work>
     </works>
 
@@ -44,6 +44,15 @@
         </transactiontype>
         <transactiontype>
             <name>GetItemByPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>UpdateItemPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>DeleteItem</name>
+        </transactiontype>
+        <transactiontype>
+            <name>InsertItem</name>
         </transactiontype>
     </transactiontypes>
 </parameters>
--- a/data/templated/example.xml
+++ b/data/templated/example.xml
@@ -70,4 +70,30 @@
          <value>2.1</value>
       </values>
    </template>
+   <template name="DeleteItem">
+      <query><![CDATA[DELETE FROM item WHERE i_price > ?]]></query>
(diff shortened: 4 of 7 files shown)


src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,9 +36,16 @@ public abstract class GenericQuery extends Procedure {
 
     /** Execution method with parameters. */
     public void run(Connection conn, List<Object> params) throws SQLException {
-        try (PreparedStatement stmt = getStatement(conn, params); ResultSet rs = stmt.executeQuery()) {
-            while (rs.next()) {
-                //do nothing
+        PreparedStatement stmt = getStatement(conn, params);
```
anjagruenheid, line 39: I would leave the creation of the stmt object in a try clause, this avoids the object being used for any other purpose (down the line).

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,9 +36,16 @@ public abstract class GenericQuery extends Procedure {
...
+        String queryString = stmt.toString().toLowerCase();
+        if (queryString.startsWith("update") || queryString.startsWith("delete") || queryString.startsWith("insert")) {
```
anjagruenheid, line 41: Is the problem that this type of statement doesn't produce a ResultSet? If so, a try/catch solution might be easier than filtering for specific strings. I.e., try executeQuery first, if an exception is thrown, try execute (I think that would return just a boolean true/false whether the query executed successfully). If that's true, yey!, if not, throw an error?

config/postgres/sample_templated_config.xml
```diff
@@ -24,7 +24,7 @@
         <work>
             <time>10</time>
             <rate>100</rate>
-            <weights>30,20,10,30,10</weights>
+            <weights>30,20,10,25,10,5</weights>
```
anjagruenheid, line 27: do at least 10, because otherwise it might not be triggered

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,21 +36,36 @@ public abstract class GenericQuery extends Procedure {
...
+            try (ResultSet rs = stmt.executeQuery()) {
+                while (rs.next()) {
+                    // do nothing
+                }
+            } catch (Exception e) {
+                try {
+                    stmt.execute();
+                } catch (Exception noExecute) {
+                    noExecute.printStackTrace();
+                    throw new RuntimeException("Not able to execute query/update: " + stmt.toString());
```
anjagruenheid, line 50: Nit: Just 'query' instead of query/update.

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,21 +36,36 @@ public abstract class GenericQuery extends Procedure {
...
-        try (PreparedStatement stmt = getStatement(conn, params); ResultSet rs = stmt.executeQuery()) {
-            while (rs.next()) {
-                //do nothing
+
+        try (PreparedStatement stmt = getStatement(conn, params)) {
+            try (ResultSet rs = stmt.executeQuery()) {
+                while (rs.next()) {
+                    // do nothing
+                }
+            } catch (Exception e) {
```
anjagruenheid, line 45: Add a short comment why we do the catch and retry.

anjagruenheid (commented):
LGTM, make it public and ask Brian for a review.

ETHenzlere:
@bpkroth could you give me a review for this PR?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,21 +36,37 @@ public abstract class GenericQuery extends Procedure {
 
     /** Execution method with parameters. */
     public void run(Connection conn, List<Object> params) throws SQLException {
-        try (PreparedStatement stmt = getStatement(conn, params); ResultSet rs = stmt.executeQuery()) {
-            while (rs.next()) {
-                //do nothing
+
+        try (PreparedStatement stmt = getStatement(conn, params)) {
+            try (ResultSet rs = stmt.executeQuery()) {
```
bpkroth, line 41: Can you restructure this to something like this instead:
https://stackoverflow.com/a/70024146
Basically, use [`execute()`](https://docs.oracle.com/javase/8/docs/api/java/sql/PreparedStatement.html#execute--) to explicitly check if it should have had a result set and then handle each case appropriately rather than implicitly relying on `executeQuery()` to throw an exception when it encounters an insert/update/delete statement.

data/templated/example.xml
```diff
@@ -61,4 +61,13 @@
          <value>11</value>
       </values>
    </template>
+    <template name="UpdateItemPrice">
+      <query><![CDATA[UPDATE item SET i_price = i_price + 1 WHERE i_price < ?]]></query>
+      <types>
+         <type>FLOAT</type>
```
bpkroth, line 67: Might be good to have a few other test cases:
DELETE, INSERT, different data types, etc.

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -67,14 +83,17 @@ public PreparedStatement getStatement(Connection conn, List<Object> params) thro
...
                     Object param = params.get(i);
-                    stmt.setObject(i + 1, param, Integer.parseInt(Types.class.getDeclaredField(paramsTypes[i]).get(null).toString()));
+                    stmt.setObject(i + 1, param,
+                            Integer.parseInt(Types.class.getDeclaredField(paramsTypes[i]).get(null).toString()));
                 } catch (Exception e) {
                     e.printStackTrace();
                     throw new RuntimeException(
-                        "Error when setting parameters. Parameter type: " + paramsTypes[i] + ", parameter value: " + params.get(i));
+                            "Error when setting parameters. Parameter type: " + paramsTypes[i] + ", parameter value: "
+                                    + params.get(i));
```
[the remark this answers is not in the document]
ETHenzlere: Yes, I had auto-formatting enabled. Removed it for now.

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,11 +36,32 @@ public abstract class GenericQuery extends Procedure {
...
+
+            while (true) {
+                if (hasResultSet) {
+                    try (ResultSet rs = stmt.getResultSet()) {
+                        while (rs.next()) {
+                            // do nothing
+                        }
+                    }
+                } else {
+                    if (stmt.getUpdateCount() == -1) {
```
anjagruenheid, line 53: Why do you need this additional check?

ETHenzlere: From Brian's pinned Stackoverflow post:
_"The loop is there because some database systems (e.g. SQL Server), can produce multiple result sets and multiple update counts, and this will handle all of them."_

config/postgres/sample_templated_config.xml
bpkroth, line 1: Can you please expand this change to the other DB types that provide a sample_templated_config.xml.
That way we can be sure that the CI pipelines pass for all of the affected DBs.
Thanks!

ETHenzlere: sure thing :+1:

config/mariadb/sample_templated_config.xml
```diff
@@ -18,13 +18,13 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
-    <!-- The workload -->
+        <!-- The workload -->
```
bpkroth, line 21:
```suggestion
    <!-- The workload -->
```

nit: whitespace change

config/sqlserver/sample_templated_config.xml
```diff
@@ -18,13 +18,13 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
-    <!-- The workload -->
+        <!-- The workload -->
```
bpkroth, line 21:
```suggestion
    <!-- The workload -->
```

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,11 +36,32 @@ public abstract class GenericQuery extends Procedure {
...
+            // False for UPDATE, INSERT, DELETE queries
+            boolean hasResultSet = stmt.execute();
+
+            while (true) {
+                if (hasResultSet) {
+                    try (ResultSet rs = stmt.getResultSet()) {
+                        while (rs.next()) {
+                            // do nothing
+                        }
+                    }
```
bpkroth, line 51: Missing `catch`?

src/main/java/com/oltpbenchmark/benchmarks/templated/procedures/GenericQuery.java
```diff
@@ -36,11 +36,32 @@ public abstract class GenericQuery extends Procedure {
...
-        try (PreparedStatement stmt = getStatement(conn, params); ResultSet rs = stmt.executeQuery()) {
-            while (rs.next()) {
-                //do nothing
+
+        try (PreparedStatement stmt = getStatement(conn, params)) {
+
+            // False for UPDATE, INSERT, DELETE queries
+            boolean hasResultSet = stmt.execute();
+
+            while (true) {
```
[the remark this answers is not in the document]
bpkroth: Maybe just add an assert inside the INSERT/UPDATE/DELETE branch that it doesn't have more results.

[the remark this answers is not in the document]
ETHenzlere: The way I understand it, we can theoretically have a statement with multiple queries (select * from x; select * from y) which would return multiple result sets. And they could theoretically also get mixed with "update" type queries.
The question is whether we want to only support this for "select" type queries, all possible combinations or we do not want to support this at all

Issue #395, opened 2023-11-20
Consider using Java >= 21 to allow usage of virtual threads


eivanov89:
Hi,

We did a [fork](https://github.com/ydb-platform/tpcc) of benchbase to get TPC-C for YDB. [Here](https://blog.ydb.tech/ydb-meets-tpc-c-distributed-transactions-performance-now-revealed-42f1ed44bd73) we describe our testing setup and TPC-C implementation with a particular focus on the threading model.

We need to run 15K TPC-C warehouses and even more, which requires >= 150,000 terminals. Currently, you use the model where 1 terminal equals 1 thread, which is not optimal. Even with sleeping threads, it's hard to create more than 3-5K terminals per server. As of Java 21, virtual threads are available, and they are in many ways similar to goroutines. Transitioning to virtual threads involves straightforward code changes [here](https://github.com/ydb-platform/tpcc/commit/1e6030cdb7c64f18d2a935dd27ead646267c5eb7). However, there might be some potential deadlocks depending on the particular JDBC driver and benchmark. For instance, when we added the c3p0 session manager, we encountered a deadlock. It was holding carrier threads waiting for sessions to become available, while session threads were parked for I/O and couldn't get carrier threads to perform I/O.

Is there any chance you would consider switching to Java 21 so that virtual threads become an option?

bpkroth:
Hi, happy to accept a PR (even partial to get started with) if you have one.  Thanks!

eivanov89:
Hi Brian, I've created the pull request. The changes are straightforward. Looking forward the review.

bpkroth:
FWIW, we bumped to java 21 a while back (and 23 today), but as I recall, there were some issues with switching to virtual threads.

I'll leave this open, but I'm going to check the title.

eivanov89:
> FWIW, we bumped to java 21 a while back (and 23 today), but as I recall, there were some issues with switching to virtual threads.
> 
> I'll leave this open, but I'm going to check the title.

Hi Brian,

Thank you for the update. Unfortunately, we (YDB) failed to allocate time to prepare a PR to the Benchbase repo. For the record:
1. We ended with two separate repos (TPC-C only): [YDB](https://github.com/ydb-platform/tpcc) and [PostgreSQL](https://github.com/ydb-platform/tpcc-postgres).
2. The commits for Postgres might be reused by Benchbase.

[...]

PR #398, opened 2023-11-26
Java 21 virtual threads

eivanov89:
This patch adds an option "vt" to use virtual threads instead of real OS threads (#395).

### DIFF
--- a/src/main/java/com/oltpbenchmark/DBWorkload.java
+++ b/src/main/java/com/oltpbenchmark/DBWorkload.java
@@ -85,6 +85,11 @@ public class DBWorkload {
       intervalMonitor = Integer.parseInt(argsLine.getOptionValue("im"));
     }
 
+    Boolean useVirtualThreads = false;
+    if (argsLine.hasOption("vt")) {
+      useVirtualThreads = Boolean.parseBoolean(argsLine.getOptionValue("vt"));
+    }
+
     // -------------------------------------------------------------------
     // GET PLUGIN LIST
     // -------------------------------------------------------------------
@@ -518,7 +523,7 @@ public class DBWorkload {
     if (isBooleanOptionSet(argsLine, "execute")) {
       // Bombs away!
       try {
-        Results r = runWorkload(benchList, intervalMonitor);
+        Results r = runWorkload(benchList, intervalMonitor, useVirtualThreads);
         writeOutputs(r, activeTXTypes, argsLine, xmlConfig);
         writeHistograms(r);
 
@@ -567,6 +572,7 @@ public class DBWorkload {
         "Base directory for the result files, default is current directory");
     options.addOption(null, "dialects-export", true, "Export benchmark SQL to a dialects file");
     options.addOption("jh", "json-histograms", true, "Export histograms to JSON file");
+    options.addOption("vt", "virtual-threads", true, "Use virtual threads instead of real threads");
     return options;
   }
 
@@ -733,7 +739,8 @@ public class DBWorkload {
     bench.loadDatabase();
   }
 
-  private static Results runWorkload(List<BenchmarkModule> benchList, int intervalMonitor)
+  private static Results runWorkload(
+      List<BenchmarkModule> benchList, int intervalMonitor, Boolean useVirtualThreads)
       throws IOException {
     List<Worker<?>> workers = new ArrayList<>();
     List<WorkloadConfiguration> workConfs = new ArrayList<>();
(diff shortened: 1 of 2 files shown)


bpkroth:
Looks like this needs to come after #399 as well.

pom.xml
```diff
@@ -12,9 +12,9 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-        <java.version>17</java.version>
-        <maven.compiler.source>17</maven.compiler.source>
-        <maven.compiler.target>17</maven.compiler.target>
+        <java.version>21</java.version>
+        <maven.compiler.source>21</maven.compiler.source>
+        <maven.compiler.target>21</maven.compiler.target>
```
bpkroth, line 17: Is it possible to separate this part (switch to java 21) to its own PR that can go first?

The docker image files would need to change at the very least, and I think I caught in a different PR/Issue discussion about other code that also needs fixing.

Thanks!

bpkroth: I started this in #410.

Let's get that merged first and then we can merge yours.

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -87,6 +87,11 @@ public static void main(String[] args) throws Exception {
             intervalMonitor = Integer.parseInt(argsLine.getOptionValue("im"));
         }
 
+        Boolean useVirtualThreads = false;
+        if (argsLine.hasOption("vt")) {
```
bpkroth, line 91: Can you add a long form `--virtual-threads` option as well?

Please also make sure the `--help` output is updated.

pom.xml
bpkroth: Did that.

src/main/java/com/oltpbenchmark/DBWorkload.java
eivanov89: Ouch, I forgot to add the option at all.

```
usage: benchbase
...
 -vt,--virtual-threads <arg>    Use virtual threads instead of real
                                threads
```

bpkroth:
@eivanov89 #416 and others made a bunch of changes to the repo at large.  Can you please rebase against and updated `main` and resolve the conflicts?

bpkroth:
> @bpkroth sure, pushed the rebased branch.

Looks like it's still got conflicts listed.

Also, can you please think about a way to add tests for this feature?

eivanov89:
> > @bpkroth sure, pushed the rebased branch.
> 
> Looks like it's still got conflicts listed.

I'm sorry for this. Forgot that my main is fork's main :) This time, did the proper rebase.

> Also, can you please think about a way to add tests for this feature?

Unfortunately I don't have enough time to work on this further. Also, the feature depends on DB client implementation. For example, I use postgres + c3p0 patch + vthreads. There was a deadlock:
* we get session for each TPC-C transaction.
* c3p0 has a session pool limited in size
* session holders do network I/O. And network I/O frees carrier thread until virtual thread is ready to execute
* when there're no sessions available, c3p0 calls `Object.wait()`, which blocks carrier thread
* benchbase ended in situation, when all carrier threads are blocked by `Object.wait()`

Your vanilla postgres impl should work though. I fixed our version by adding a sessions semaphore (waiting on it doesn't block carrier thread), so that c3p0 never calls `Object.wait()`. I'll check if I can bring patch with c3p0 and the semaphore. c3p0 looks like a very useful addition.

bpkroth:
> > > @bpkroth sure, pushed the rebased branch.
> > 
> > 
> > Looks like it's still got conflicts listed.
> 
> I'm sorry for this. Forgot that my main is fork's main :) This time, did the proper rebase.
> 
> > Also, can you please think about a way to add tests for this feature?
> 
> Unfortunately I don't have enough time to work on this further. Also, the feature depends on DB client implementation. For example, I use postgres + c3p0 patch + vthreads. There was a deadlock:
> 
> * we get session for each TPC-C transaction.
> * c3p0 has a session pool limited in size
> * session holders do network I/O. And network I/O frees carrier thread until virtual thread is ready to execute
> * when there're no sessions available, c3p0 calls `Object.wait()`, which blocks carrier thread
> * benchbase ended in situation, when all carrier threads are blocked by `Object.wait()`
> 
> Your vanilla postgres impl should work though. I fixed our version by adding a sessions semaphore (waiting on it doesn't block carrier thread), so that c3p0 never calls `Object.wait()`. I'll check if I can bring patch with c3p0 and the semaphore. c3p0 looks like a very useful addition.

Sounds good.  We're happy to take a PR for both, but it's always good to have a test associated with new features to make sure we continue to support them in the future and they don't break existing functionality.

eivanov89:
I had missed the point, that there are a lot of synchronized inside the benchmark, which should be rewritten at first.

Issue #369, opened 2023-09-29
GetInterfaces test in TestClassUtil.java appears to be checking interfaces of Java builtins?


josh-thisisbud:
I just want to prefix all of this with I know very little about Java, and I'm only building this tool because I'd like to use it to benchmark, having pre-compiled versions would be really handy.

I modified the test to:
```
    /**
     * GetInterfaces
     */
    @Test
    public void testGetInterfaces() {
        Class<?>[] expected = {
                Serializable.class,
                Cloneable.class,
                Iterable.class,
                Collection.class,
                List.class,
                RandomAccess.class,
        };
        Collection<Class<?>> results = ClassUtil.getInterfaces(target_class);
        // System.err.println(target_class + " => " + results);
        assert (!results.isEmpty());
        assertEquals(String.format("Target class was %s", target_class), expected.length, results.size());

        for (Class<?> e : expected) {
            assert (results.contains(e));
        } // FOR
    }
```

This resulted in output of

```
testGetInterfaces(com.oltpbenchmark.util.TestClassUtil)  Time elapsed: 0.003 sec  <<< FAILURE!
java.lang.AssertionError: Target class was class java.util.ArrayList expected:<6> but was:<7>
        at org.junit.Assert.fail(Assert.java:89)
        at org.junit.Assert.failNotEquals(Assert.java:835)
        at org.junit.Assert.assertEquals(Assert.java:647)
        at com.oltpbenchmark.util.TestClassUtil.testGetInterfaces(TestClassUtil.java:130)
        at java.base/jdk.internal.reflect.DirectMethodHandleAccessor.invoke(DirectMethodHandleAccessor.java:103)
        at java.base/java.lang.reflect.Method.invoke(Method.java:580)
        at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:59)
        at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
        at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:56)
        at org.junit.internal.runners.statements.InvokeMethod.evaluate(InvokeMethod.java:17)
        at org.junit.runners.ParentRunner$3.evaluate(ParentRunner.java:306)
        at org.junit.runners.BlockJUnit4ClassRunner$1.evaluate(BlockJUnit4ClassRunner.java:100)
        at org.junit.runners.ParentRunner.runLeaf(ParentRunner.java:366)
[...]

PR #410, opened 2023-12-07
Upgrade to Java 21

bpkroth:
Separates a conversion to Java 21 from #398 and also handles errors encountered in #369.


bpkroth:
> @apavlo staging this as a smaller change, but there's a few other related ones. Main concern here is that it is potentially a breaking change for folks using just the jar release, so we should probably also tag a new release commit once this is merged. This change also incorporates a new version tag in the relevant files to match that. Let me know if you have any other thoughts.

Discussed offline.  Gonna merge this now and tag a new release.  Thanks!

Issue #406, opened 2023-12-06
Add linting and formatting pre-commit/CI checks


bpkroth:
This is more a convenience to developers and reviewers.

Often changes include some tweaks from an IDE's opinion on reformatting that reviewers have to adjust or ask for revisions on.

We can (should) automate this with pre-commit hooks and CI checks to remove a burden on both developers and reviewers.

Introducing these changes however will likely reformat a large portion of the codebase, so should be kept to a separate PR.

PR #413, opened 2023-12-08
Enable strict compiler settings

bpkroth:
This change enforces javac compiler warnings and linting as errors as a part of #406.

Splitting work out from #410, though that should be merged first, and may require additional fixups after that (e.g., new compiler constraints).


src/main/java/com/oltpbenchmark/benchmarks/wikipedia/data/RevisionHistograms.java
```diff
@@ -290,12 +290,16 @@ public abstract class RevisionHistograms {
      */
     public static final int[] REVISION_DELTA_SIZES = {1000, 10000, 100000};
 
+    /**
+     * Simple generic class overload to avoid some cast warnings below.
```
anjagruenheid, line 294: Should this type of handling be part of a best practices doc? I.e., avoid casting, use some structures like this instead.

src/test/java/com/oltpbenchmark/benchmarks/auctionmark/TestAuctionMarkWorker.java
```diff
@@ -61,7 +60,7 @@ public void testUniqueSellers() throws Exception {
         Set<UserId> all_users = new HashSet<UserId>();
         Set<UserId> worker_users = new TreeSet<UserId>();
         Integer last_num_users = null;
-        for (Worker w : this.workers) {
+        for (var w : this.workers) {
```
anjagruenheid, line 63: Why the change?

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkLoader.java
```diff
@@ -1195,6 +1195,7 @@ protected class ItemBidGenerator extends SubTableGenerator<LoaderItemInfo> {
         private LoaderItemInfo.Bid bid = null;
         private float currentBidPriceAdvanceStep;
         private long currentCreateDateAdvanceStep;
+        @SuppressWarnings("unused") // spurious
```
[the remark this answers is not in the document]
bpkroth: Without it `currentPrice` get's an alert about being unused (it's only ever assigned, not read/returned).

src/main/java/com/oltpbenchmark/benchmarks/wikipedia/data/RevisionHistograms.java
bpkroth: Not a bad idea.  It's basically a limitation in the java language spec as I understand it.  You can't create arrays of a generics:
https://docs.oracle.com/javase/tutorial/java/generics/restrictions.html#createArrays

src/test/java/com/oltpbenchmark/benchmarks/auctionmark/TestAuctionMarkWorker.java
bpkroth: TL;DR: it made the error go away 😅

`this.workers` returns a `List<Worker<? extends BenchmarkModule>>` (i.e., a wildcard for any number of classes that extend `BenchmarkModule`).
Critically, it doesn't return `List<Workers<T extends BenchmarkModule>>`.
That means that `Worker w : this.workers` is technically an invalid cast.
But so is `Worker<AuctionMarkBenchmark>` even though that's what's actually in the list, despite what the type declaration says.
The problem is that the `this.workers` value actually gets populated from an overload of `BenchmarkModule` baseclass, `AuctionMarkBenchmark`, but that means that we'd wind up with a circular type definition (where the `AuctionMarkBenchmark extends BenchmarkModule` needs to return a `Worker<AuctionMarkBenchmark>`).
Anyways, that is quite complex so I wanted to push it off to another PR (if at all) and it turns out that by simply letting java infer the type via a `var` declaration, we avoid the issue (for now).
We're still stuck with the explicit cast after that though unfortunately.

src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
```diff
@@ -77,14 +77,15 @@ public abstract class BenchmarkModule {
     public BenchmarkModule(WorkloadConfiguration workConf) {
         this.workConf = workConf;
         this.dialects = new StatementDialects(workConf);
-        initClassLoader();
+        // setClassLoader();
+        this.classLoader = ClassLoader.getSystemClassLoader();
```
bpkroth, line 81: Adjusted to avoid a `this-escape` warning.

src/main/java/com/oltpbenchmark/benchmarks/auctionmark/AuctionMarkLoader.java
```diff
@@ -46,7 +46,7 @@
  * @author pavlo
  * @author visawee
  */
-public class AuctionMarkLoader extends Loader<AuctionMarkBenchmark> {
+public final class AuctionMarkLoader extends Loader<AuctionMarkBenchmark> {
```
bpkroth, line 49: Many of these classes were converted to `final` to  avoid `this-escape` warnings in which non-final methods are used by the class to initialize the class.  In theory those could be overridden by a subclass to result in problematic initialization.

Since most of these classes are never expected to be subclassed, we simply avoid the error for now and mark them as `final`.

bpkroth: Added notes to this effect in the CONTRIBUTING.md file.

PR #427, opened 2023-12-14
Improve test result checking

bpkroth:
- Add checks on the result outputs to make sure that the length of time and number of queries expected to run were.
- Make sure to exit non-zero when RuntimeExceptions are encountered.
- Temporarily remove a faulty INSERT test from the templated benchmark.

Split out from #283.


data/templated/example.xml
```diff
@@ -79,6 +79,7 @@
          <value>255.0</value>
       </values>
    </template>
+   <!-- FIXME: This results in primary key failures due to duplicate keys.
```
bpkroth, line 82: Temporarily removes a faulty INSERT test query from the templated benchmark example introduced in #386 since it results in duplicate keys.

config/oracle/sample_templated_config.xml
```diff
@@ -45,5 +45,16 @@
         <transactiontype>
             <name>GetItemByPrice</name>
         </transactiontype>
+        <transactiontype>
+            <name>UpdateItemPrice</name>
+        </transactiontype>
+        <transactiontype>
+            <name>DeleteItem</name>
+        </transactiontype>
```
bpkroth, line 53: Oracle tests were missing a few items.

PR #445, opened 2024-01-03
Bug fix: Thread.sleep required long type (in milliseconds) not Duration

ranaalotaibiMS:
Thread.sleep required long type (in milliseconds) not Duration
CC: @bpkroth


ranaalotaibiMS:
Java 21 was needed!

Issue #428, opened 2023-12-14
Fix faulty INSERT test query from the templated benchmark example introduced in #386 since it results in duplicate keys


bpkroth:
Need to fix INSERT test query from the templated benchmark example introduced in #386 since it results in duplicate keys.

_Originally posted by @bpkroth in https://github.com/cmu-db/benchbase/pull/427#discussion_r1427201975_

bpkroth:
@ETHenzlere can you please take a look at this?

bpkroth:
@anjagruenheid for visibility

bpkroth:
See Also: https://github.com/cmu-db/benchbase/pull/427/commits/7d4e90b2b6fd72eaf86dc4669da594768a7ff8c0

PR #447, opened 2024-01-05
Fix: Changed INSERT statements of templated benchmarks

ETHenzlere:
INSERT query was not working due to duplicate primary keys

Simple solutions on the query did not work for all systems
(For example _INSERT IGNORE, INSERT -- IF NOT EXISTS_).

For now I changed the table of the insert to one that does not have a primary key.

This PR tries to solve #428


bpkroth (approved):
LGTM, thanks!

PR #455, opened 2024-01-24
Feat: Add Skeleton to support Dataset anonymization with differential privacy

ETHenzlere:
**DP-Anonymization for BenchBase**

In order to measure the influence of modern anonymization techniques on datasets and their query performance, we would like to add support for differential privacy mechanisms to BenchBase. In this PR, we have added the skeleton of our approach to the codebase. We would like to introduce a new flag: `--anonymize=true` to the system that allows to anonymize tables after the loading step and before execution.

The anonymization will:

1. Pull data from the DBMS via already specified JDBC connection (config)
2. Run DP-algorithms on the data to create a new, synthetic dataset
3. Push the synthetic data back to the DBMS as an anonymized copy

The anonymization information must be provided in the config file. The process will work with minimal information but also allow for fine-tuning. A separate README file has been constructed that will list all the features and how to use them.
`/scripts/anonymization/README.md`

Minimal config:
```
 <anonymization>
        <table name="item">
          <differential_privacy />
        </table>
   </anonymization>
```
Sensitive value handling is one feature we want to add to the process immediately. It replaces actual values of specified columns with fake ones. The code base has already been written, tested and used privately within BenchBase. 

The column faking approach will be decoupled from differential privacy, to allow for more control.
```
 <anonymization>
        <table name="item">
          <differential_privacy> ... </differential_privacy>
          <value_faking> ... </value_faking>
        </table>
   </anonymization>
```

**Disclaimer: The anonymization itself is not part of this PR in order to reduce the complexity. Currently, the anonymization flag will call the script and parse the config.  The rest of the code is ready-to-be-added**

[Architecture Benchbase.pdf](https://github.com/cmu-db/benchbase/files/14356135/Architecture.Benchbase.pdf)


config/postgres/sample_templated_config.xml
```diff
@@ -19,6 +19,22 @@
 
     <query_templates_file>data/templated/example.xml</query_templates_file>
 
+    <anon>
```
anjagruenheid, line 22: Do you expect anything else to be part of anon? Otherwise, we might be able to drop the anonTable tag?

config/postgres/sample_templated_config.xml
```diff
@@ -19,6 +19,22 @@
...
+        <anonTable>
+            <tablename>item</tablename>
+            <droppable>i_id</droppable>
+            <categorical>i_name,i_data,i_im_id</categorical>
+            <continuous>i_price</continuous>
+            <ordinal></ordinal>
+            <eps>1.0</eps>
+            <preEps>0.5</preEps>
```
anjagruenheid, line 30: I think the standard for xml is to use "_" instead of capitalization.

config/postgres/sample_templated_config.xml
```diff
@@ -19,6 +19,22 @@
...
+            <algorithm>aim</algorithm>
```
anjagruenheid, line 31: You need to document somewhere what the available options are.

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -775,4 +789,39 @@ private static boolean isBooleanOptionSet(CommandLine argsLine, String key) {
...
+    }
+
+    LOG.info("Starting the Anonymization process");
+    LOG.info(SINGLE_LINE);
+    ProcessBuilder processBuilder =
+        new ProcessBuilder("python3", "scripts/anonymizer.py", configFile, templatesPath);
+    try {
+      // Redirect Output stream of the script to get live feedback
+      processBuilder.inheritIO();
+      Process process = processBuilder.start();
```
anjagruenheid, line 815: For a stub, you should add what's expected to be returned right now and a short disclaimer that this will be extended in the future.

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -775,4 +789,39 @@ private static boolean isBooleanOptionSet(CommandLine argsLine, String key) {
...
+      int exitCode = process.waitFor();
+      if (exitCode != 0) {
+        throw new Exception("Anonymization program exited with a non-zero status code");
+      }
+      LOG.info("Finished the Anonymization process for all tables");
```
anjagruenheid, line 820: Not true right now, since you're only parsing the config?

scripts/anonymizer.py
```diff
@@ -0,0 +1,189 @@
+"""Module that handles the full Anonymization pipeline
+"""
+import sys
+import xml.etree.ElementTree as ET
+
+def rewriteFakeQueries(fakeValues, path):
```
anjagruenheid, line 6: I think that the API currently lacks context. Maybe add a short flow description in the overview/main somewhere to explain why you need these specific functions.

config/postgres/sample_templated_config.xml
ETHenzlere: It is used to handle multiple tables at once. I have renamed it for now.

config/postgres/sample_templated_config.xml
```diff
@@ -19,6 +19,24 @@
...
 
+    <!-- The anonymization configuration -->
+    <anonymization>
+        <table name="item" epsilon="1.0" pre_epsilon="0.0" algorithm="aim">
+            <!-- Column categorization -->
+            <droppable>i_id</droppable>
+            <categorical>i_name,i_data,i_im_id</categorical>
+            <continuous>i_price</continuous>
+            <!-- Continuous column fine-tuning -->
+            <continuousConfig>
```
bpkroth, line 30:
```suggestion
            <continuous_config>
```

nit: we have a mix of camel and underscore in here atm I know, but I think it's just one case which we can separately fix, so let's not make that worse here.

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
+# Anonymization
+
+<span style="color:orange"> ANONYMIZATION IS A WORK IN PROGRESS AND DOES CURRENTLY NOT ACTUALLY ANONYMIZE THE DATA. THIS FEATURE WILL BE ADDED LATER </span>
+
+The anonymization module allows to apply privacy mechanisms such as differential privacy or column faking to the data.
```
bpkroth, line 5:
```suggestion
The anonymization module allows applying privacy mechanisms such as differential privacy or column faking to the data.
```

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
+# Anonymization
+
+<span style="color:orange"> ANONYMIZATION IS A WORK IN PROGRESS AND DOES CURRENTLY NOT ACTUALLY ANONYMIZE THE DATA. THIS FEATURE WILL BE ADDED LATER </span>
```
bpkroth, line 3:
```suggestion
*NOTE: ANONYMIZATION IS A WORK IN PROGRESS AND DOES CURRENTLY NOT ACTUALLY ANONYMIZE THE DATA. THIS FEATURE WILL BE ADDED LATER*
```

scripts/anonymization/README.md
bpkroth: Maybe also add links to external reference sources on differential privacy.

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
...
+        algorithm="patectgan"
+    />
+</anonymization>
+```
+
+### Column Information
+
+The columns of a table can be split into four categories:
+
+1. `drop` - Columns that are omitted from the DP-mechanism
```
bpkroth, line 36: Clarifying: 
`drop` -> "omitted" or "removed"?

i.e., are those columns skipped when applying differential privacy but still exist in the dataset at the end as they originally appeared,
or are the "removed" from the dataset (i.e, `ALTER TABLE DROP COLUMN ...`)

bpkroth: If the former, maybe call it something else like "passthru" to avoid confusion?

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+```xml
+<anonymization>
+    <table name="item">
+        <categorical>i_name,i_data,i_im_id</categorical>
+        <continuous>i_price</continuous>
+        <ordinal>i_id</ordinal>
+    </table>
+</anonymization>
+```
```
bpkroth, line 71: Alternative XML schema:

```xml
<anonymization>
  <table name="t">
    <columns>
      <column name="c1" type="continuous" />
      <column name="c2" type="continuous" />
      <column name="c2" type="ordinal" />
      <column name="c3" type="categorical" />
      <column name="c4" type="passthru" />
  </table>
```

bpkroth: Basically, avoid sub parsing in another language (csv) and just use xml for the whole thing in some way.

bpkroth: Also, then all your "fine tuning" configs below can get merged into this rather than having two separate sections.

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
...
+        <continuousConfig>
+            <column name="i_price" bins="1000" lower="2.0" upper="100.0">
+        <continuousConfig>
+    </table>
+</anonymization>
+```
+
+### Sensitive Values
+
+The anonymization module supports value faking to handle sensitive values. A `<sensitive>`-tag must be created to signify that column faking should be applied. Column faking can be defined individually on each column. To do so, parameters can be added inside of the `<column>`-tag. (See below)
```
bpkroth, line 117:
```suggestion
The anonymization module supports value faking to handle sensitive values.
A `<sensitive>`-tag must be created to signify that column faking should be applied.
Column faking can be defined individually on each column.
To do so, parameters can be added inside of the `<column>`-tag. (See below)
```

nit: one sentence per line

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
...
+| ------- | ------- | ------------------------------------------------ | ---------------------------------------------------------------------------------------------- |
+| name    | -       | Any string                                       | The name of the column                                                                         |
+| method  | pystr   | Any method provided by the python faker library  | The faking method of the library. If the method is not found, a random string will be inserted |
+| locales | -       | A string of supported locales separated by comma | The locale to produce localized values like chinese street names or english first names        |
+| seed    | 0       | Any integer value                                | The privacy budget spent on preprocessing of the data. Only necessary for continuous columns   |
+
+```xml
+<anonymization>
+    <table name="item" algorithm="aim">
+        <sensitive>
```
bpkroth, line 129: Is "column faking" mutually exclusive with "differential privacy"?

Or can the schema somehow be unified with the one I mentioned above?

scripts/anonymization/README.md
```diff
@@ -0,0 +1,220 @@
...
+        </sensitive>
+    </table>
+</anonymization>
+```
+
+## Examples
+
+### Basic
+
+The most basic config will need only the name of the table and apply default values for the anonymization.
```
bpkroth, line 146: Document current defaults, or where to look them up?

src/main/java/com/oltpbenchmark/DBWorkload.java
```diff
@@ -775,4 +791,40 @@ private static boolean isBooleanOptionSet(CommandLine argsLine, String key) {
...
+    String templatesPath = "";
+    if (xmlConfig.containsKey("query_templates_file")) {
+      templatesPath = xmlConfig.getString("query_templates_file");
+    }
+
+    LOG.info("Starting the Anonymization process");
+    LOG.info(SINGLE_LINE);
+    ProcessBuilder processBuilder =
+        new ProcessBuilder(
+            "python3", "scripts/anonymization/anonymizer.py", configFile, templatesPath);
```
bpkroth, line 814: Why python and not java?
Alt, why does this need to be invoked directly in benchbase at all?

For instance, the following shell could suffice, no?

```sh
java -jar benchbase.jar --load=true ...
anononymizer.py ...
java -jar benchbase.jar --execute=true ...
```

scripts/anonymization/README.md
ETHenzlere: renamed to "ignore". Description is in the README. Ignored columns are not anonymized but added back in their original form to the DB

scripts/anonymization/README.md
ETHenzlere: Changed Schema to avoid sub parsing. See README

scripts/anonymization/README.md
ETHenzlere: Changed this part of the README. Automatic anonymization will automatically try to categorize columns. 

Defaults are documented and I provided a link to the algorithm repository as additional info

src/main/java/com/oltpbenchmark/DBWorkload.java
ETHenzlere: Python because the algorithms are only available in Python. 

It does not need to be invoked directly (You can run it as a standalone) but we thought it would be nice to have since there is information from the config file of BenchBase the script re-uses (JDBC connection info)

scripts/anonymization/README.md
ETHenzlere: We would like to handle it separately as part of the Anonymization.

This allows for anonymization of sensitive columns without differential privacy (For example for the Twitter workload, user_profiles table)

or coupled anonymization (Run DP first and then handle the sensitive values for higher privacy guarantees)

scripts/anonymization/README.md
```diff
@@ -0,0 +1,322 @@
...
+        </differential_privacy>
+    </table>
+</anonymization>
+```
+
+### Column Information
+
+The columns of a table can be split into four categories:
+
+1. `ignore` - Columns that are ignored by the DP-mechanism. Left untouched
```
bpkroth, line 53:
```suggestion
1. `ignore` - Columns that are ignored by the DP-mechanism. Left untouched.
```

nit: consistency about ending sentences of list items in `.` periods.

scripts/anonymization/README.md
```diff
@@ -0,0 +1,322 @@
...
+            </continuous>
+            <ordinal>
+                <column name="i_id"/>
+            </ordinal>
+        </differential_privacy>
+    </table>
+</anonymization>
+```
+
+<span style="color:red"> **Disclaimer**: As soon as the column types are defined by hand, all columns must be defined. It is not possible to only specify some of the categorical columns and let the algorithms do the rest!</span>
```
bpkroth, line 109: Question about this: Couldn't you
1) start with inferred/discovered defaults
2) use any additional column details in the config to override those defaults
so that you wouldn't have to either specify no columns or specify all columns?

scripts/anonymization/anonymizer.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    # TODO
+
+
+def getTransformer(
+    dataset, algorithm, categorical, continuous, ordinal, continuousConfig
+):
+    """Function that creates a custom transformer for the dp-mechanism
+    """
+
```
bpkroth, line 49:
```suggestion
```
nit: inconsistent whitespace

scripts/anonymization/anonymizer.py
```diff
@@ -0,0 +1,210 @@
...
+    """
+    # TODO
+
+
+def anonymize(
+    dataset: str,
+    anon_config: dict,
+    sens_config: list,
+    cont_config: list,
+    templatesPath: str,
```
bpkroth, line 64: Is it possible to set typing hints for all of the functions?  Or leave that until you fill in the details later on?

config/postgres/sample_templated_config.xml
bpkroth, line 1: Will eventually probably want to ensure we have additional DB backend support, at least the common ones if possible (e.g., MySQL, MariaDB, SqlServer).

bpkroth, line 1: May also want to separate this to its own config file so we can independently test:
a) template benchmark
b) anonymization

bpkroth: And speaking of tests, I know this isn't supposed to do anything yet, but please also add some test cases yet to validate that.
They can always be expanded on later on.

.github/workflows/python.yml
```diff
@@ -0,0 +1,31 @@
+name: Test Anonymization (Python)
```
bpkroth, line 1:
```suggestion
name: BenchBase (Python)
```

scripts/anonymization/README.md
bpkroth, line 1: Please add a short section on setup/requirements (e.g., python 3.10, pip install requirements.txt, etc.) and running tests locally (e.g., pytest)

.github/workflows/python.yml
```diff
@@ -0,0 +1,31 @@
...
+      - name: Set up Python ${{ matrix.python-version }}  
+        uses: actions/setup-python@v4  
+        with:  
+          python-version: ${{ matrix.python-version }}  
+      - name: Install dependencies  
+        working-directory: ./scripts/anonymization
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt  
+      - name: Test with pytest
```
bpkroth, line 29:
```suggestion
      - name: Test anonymization with pytest 
```

That way we can add others in the future.

.github/workflows/python.yml
```diff
@@ -0,0 +1,31 @@
...
+        with:  
+          python-version: ${{ matrix.python-version }}  
+      - name: Install dependencies  
+        working-directory: ./scripts/anonymization
+        run: |
+          python -m pip install --upgrade pip
+          pip install -r requirements.txt  
+      - name: Test with pytest 
+        working-directory: ./scripts/anonymization 
+        run: pytest test.py
```
bpkroth, line 31: Can you please add some sort of basic linter and style check as well?  For instance, `flake8` and/or `pylint`?

scripts/anonymization/README.md
ETHenzlere: Unfortunately, this is the architecture of the DP-algorithms library that we use. (They do not allow mixed configurations)

.github/workflows/python.yml
ETHenzlere: Added pylint

scripts/anonymization/.pylintrc
bpkroth, line 1: Can probably move this up a few levels so it covers other python scripts in the future as well.

Also, can you leave it as just the overrides, and not the full config?

`--fail-under=90` can get moved here for instance instead.

.github/workflows/maven.yml
```diff
@@ -286,6 +300,11 @@ jobs:
           if [[ ${{matrix.benchmark}} == templated ]]; then
             java -jar benchbase.jar -b tpcc -c config/mysql/sample_tpcc_config.xml --create=true --load=true --execute=false --json-histograms results/histograms.json
             java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/mysql/sample_${{matrix.benchmark}}_config.xml --create=false --load=false --execute=true --json-histograms results/histograms.json
+          # For anonymization, we load tpcc and anonymize a single table. The workload itself is not executed
+          elif [[ ${{matrix.benchmark}} == anonymization ]]; then
+            java -jar benchbase.jar -b tpcc -c config/mysql/sample_tpcc_config.xml --create=true --load=true --execute=false --json-histograms results/histograms.json
+            java -jar benchbase.jar -b tpcc -c config/mysql/sample_${{matrix.benchmark}}_config.xml --anonymize=true
+            exit 0
```
bpkroth, line 307:
```suggestion
```

.github/workflows/maven.yml
```diff
@@ -286,6 +300,11 @@ jobs:
           if [[ ${{matrix.benchmark}} == templated ]]; then
             java -jar benchbase.jar -b tpcc -c config/mysql/sample_tpcc_config.xml --create=true --load=true --execute=false --json-histograms results/histograms.json
             java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/mysql/sample_${{matrix.benchmark}}_config.xml --create=false --load=false --execute=true --json-histograms results/histograms.json
+          # For anonymization, we load tpcc and anonymize a single table. The workload itself is not executed
+          elif [[ ${{matrix.benchmark}} == anonymization ]]; then
+            java -jar benchbase.jar -b tpcc -c config/mysql/sample_tpcc_config.xml --create=true --load=true --execute=false --json-histograms results/histograms.json
+            java -jar benchbase.jar -b tpcc -c config/mysql/sample_${{matrix.benchmark}}_config.xml --anonymize=true
```
bpkroth, line 306: Why aren't we executing the workload?  Is that a future PR?

.github/workflows/maven.yml
```diff
@@ -286,6 +300,11 @@ jobs:
           if [[ ${{matrix.benchmark}} == templated ]]; then
             java -jar benchbase.jar -b tpcc -c config/mysql/sample_tpcc_config.xml --create=true --load=true --execute=false --json-histograms results/histograms.json
             java -jar benchbase.jar -b ${{matrix.benchmark}} -c config/mysql/sample_${{matrix.benchmark}}_config.xml --create=false --load=false --execute=true --json-histograms results/histograms.json
+          # For anonymization, we load tpcc and anonymize a single table. The workload itself is not executed
```
bpkroth, line 303: Does the backend matter for coverage?  Right now we're only executing this for MySQL.

.github/workflows/maven.yml
ETHenzlere: With only the skeleton, the anonymization does not do anything. Once fully implemented, it would create an anonymized **copy** of one table. The original table would then have to be replaced such that running the benchmark would make sense. For now I would suggest to leave it out

.github/workflows/maven.yml
ETHenzlere: necessary as long as no actual benchmark is run (otherwise it will always fail because it does not find any results of benchmarking runs

.github/workflows/maven.yml
ETHenzlere: Does not matter. I was waiting for feedback before supporting more DBs

.github/workflows/maven.yml
bpkroth: Please add comments to that effect.  We can remove them in the future PR when you enable the rest of the framework and tests.

.github/workflows/maven.yml
bpkroth: Understood.  I guess this goes along with the explanation in the other comment about not executing yet.  Please put a comment in for now so we know to remove this part once the execution phase is added.

bpkroth (approved):
Please add the comments I asked for and then I think we can merge this.
Thanks!

PR #466, opened 2024-02-14
Fix scan length distribution for YCSB

stonewhitener:
## TL;DR

This patch changes the distribution of scan lengths in YCSB from a Zipfian distribution to a uniform distribution.

## Background

The default distribution of scan lengths in YCSB Workload E is uniform [^1] [^2] [^3]. The current implementation uses Zipfian distribution with a specified skew factor, which should be avoided for fair evaluation of scan performance.

Screenshot 2024-02-14 at 15 48 10

[^1]: https://github.com/brianfrankcooper/YCSB/blob/ce3eb9/workloads/workloade#L44
[^2]: https://github.com/brianfrankcooper/YCSB/blob/ce3eb9/core/src/main/java/site/ycsb/workloads/CoreWorkload.java#L536
[^3]: https://doi.org/10.1145/1807128.1807152

### DIFF
--- a/src/main/java/com/oltpbenchmark/distributions/UniformGenerator.java
+++ b/src/main/java/com/oltpbenchmark/distributions/UniformGenerator.java
@@ -1,0 +1,38 @@
+package com.oltpbenchmark.distributions;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+public class UniformGenerator extends IntegerGenerator {
+  int min, max;
+
+  /**
+   * Create a uniformly distributed random number generator for items.
+   *
+   * @param items Number of items.
+   */
+  public UniformGenerator(int items) {
+    this(0, items - 1);
+  }
+
+  /**
+   * Create a uniformly distributed random number generator for items between min and max
+   * (inclusive).
+   *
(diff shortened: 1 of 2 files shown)


bpkroth (approved):
LGTM
@apavlo any thoughts?

src/main/java/com/oltpbenchmark/distributions/UniformGenerator.java
```diff
@@ -0,0 +1,38 @@
+package com.oltpbenchmark.distributions;
+
+import java.util.concurrent.ThreadLocalRandom;
+
+public class UniformGenerator extends IntegerGenerator {
+  int min, max;
```
apavlo, line 6: I realize this is trivial but split these two members across separate lines to match the coding style of other classes.

apavlo (approved):
Minor nit request.

src/main/java/com/oltpbenchmark/distributions/UniformGenerator.java
stonewhitener: Done!

PR #480, opened 2024-03-09
Let --dialects-export output sql in log

### LINKED ISSUE #252: --dialects-export option doesn't work
I tried --dialects-export option for tpcc and tpch with postgres, but it seems not work. Nothing is output on the console and no SQL file is created.

juvenn:
Respect `--dialects-export` option, when enabled, it should output SQL in log.

Close #252 

### Example

```xml
mvn compile exec:java -P postgres -Dexec.args="--dialects-export true -b twitter -c twitter_config.xml"

...
<dialects>
    <dialect type="POSTGRES">
        <procedure name="GetFollowers">
            <statement name="getFollowerNames">SELECT uid, name FROM "user_profiles" WHERE uid IN (??)</statement>
            <statement name="getFollowers">SELECT f2 FROM "followers" WHERE f1 = ? LIMIT 20</statement>
        </procedure>
        <procedure name="GetTweet">
            <statement name="getTweet">SELECT * FROM "tweets" WHERE id = ?</statement>
        </procedure>
        <procedure name="GetTweetsFromFollowing">
            <statement name="getTweets">SELECT * FROM "tweets" WHERE uid IN (??)</statement>
            <statement name="getFollowing">SELECT f2 FROM "follows" WHERE f1 = ? LIMIT 20</statement>
        </procedure>
        <procedure name="GetUserTweets">
            <statement name="getTweets">SELECT * FROM "tweets" WHERE uid = ? LIMIT 10</statement>
        </procedure>
        <procedure name="InsertTweet">
            <statement name="insertTweet">INSERT INTO "added_tweets" VALUES (default, ?, ?, ?)</statement>
        </procedure>
    </dialect>
</dialects>
```

### DIFF
--- a/src/main/java/com/oltpbenchmark/DBWorkload.java
+++ b/src/main/java/com/oltpbenchmark/DBWorkload.java
@@ -449,7 +449,7 @@ public class DBWorkload {
                 .export(
                     bench.getWorkloadConfiguration().getDatabaseType(),
                     bench.getProcedures().values());
-        LOG.debug(xml);
+        LOG.info(xml);
         System.exit(0);
       }
       throw new RuntimeException("No StatementDialects is available for " + bench);


bpkroth:
Hi, can you please add a test case to the CI pipeline for this?

juvenn:
Hi Brian, sorry for late reply. I think a better approach would be to output sql to a file, which would be easier for user to work with, and for us to test. It may warrant another MR though. Perhaps, we can get this simple fix merged before that? What do you think?

juvenn:
What would you like to test in this case?

I'm thinking about output it to a file, such as `sql-dilalects.xml`. Then assert it exists after run the command. Other than that, asserting the contents structure might be a bit stretched. What do you have in mind?

bpkroth:
> What would you like to test in this case?

I like your idea of outputting it to a different file.  Given it's broken atm anyways, I'd suggest we change the CLI arg to accept a file name to output it to, perhaps defaulting to `sql-dialects.xml`.

> I'm thinking about output it to a file, such as `sql-dilalects.xml`. Then assert it exists after run the command. Other than that, asserting the contents structure might be a bit stretched. What do you have in mind?

I think checking the contents are at least well formed and have *something* reasonable in them seems like a good move to me too.

Can probably use `xmllint` or a simple `python` script for this.  Something simple and reusable across DB types would be ideal to start with.  For instance, the check could run against a particular `--benchmark` and look for a specific number of leading keywords (e.g., 10 `SELECT`, 3 `UPDATE`, etc.) without having to be concerned with other details about the query syntax.  We should do that across the "supported" DBs that are in the `.github/workflows/maven.yml` CI pipeline.  Make sense?

Pretty sure we can do all of that in this PR pretty easily.

PR #517, opened 2024-05-20
CHBenchmark dialect-sqlserver.xml

mush-zhang:
Added Sql Server Dialect for CH-benCHmark.

### DIFF
--- a/src/main/resources/benchmarks/chbenchmark/dialect-sqlserver.xml
+++ b/src/main/resources/benchmarks/chbenchmark/dialect-sqlserver.xml
@@ -1,0 +1,231 @@
+<?xml version="1.0"?>
+<dialects>
+    <dialect type="SQLSERVER">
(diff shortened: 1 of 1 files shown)


anjagruenheid:
LGTM, can you add a @bpkroth as reviewer?

apavlo:
This looks reasonable to me. Unfortunately we don't run CH-BenCHmark in our tests.

bpkroth:
LGTM, but as @apavlo mentioned, can you please add it to CI tests here so we make sure this gets tested:
https://github.com/cmu-db/benchbase/blob/main/.github/workflows/maven.yml#L582

bpkroth:
Hi @mush-zhang, I merged #519 to enable chbenchmark for most other backends.  Can you please update your branch and enable tests following that pattern and let's make sure it passes CI so we can merge this?
Thanks!

PR #530, opened 2024-06-04
Configure healthcheck to avoid error logs

BuonHobo:
Added options to the healthcheck to avoid seeing "FATAL:  role "root" does not exist" many many times in the log. The message still appears only two times when starting the container but that's probably caused by something else.


bpkroth (approved):
Made a minor extension to this to remove the other couple of error message sources as well.  Thanks!

BuonHobo:
I'm glad that it turned out useful! Thanks!

PR #586, opened 2024-12-10
feat: tpc-c support for spanner

git-anubhav:


bpkroth:
Looks reasonable to me so far, but would be good to have spanner runs in the CI testing pipeline:
https://github.com/cmu-db/benchbase/blob/main/.github/workflows/maven.yml

bpkroth (commented):
Looks reasonable.  Could use some tests I think.

PR #594, opened 2025-02-12
Optimize TPCH DDL for faster MySQL loading

nitinbil:
This pull request includes changes to improve the performance of loading the TPCH database on a MySQL instance by deferring index creation to a post-load script. The most important changes include adding a new post-load script reference and removing index creation statements from the DDL script.

Performance improvements:

* [`config/mysql/sample_tpch_config.xml`](diffhunk://#diff-a48fec690069601e7837968f1848c04583a74352c5d98597f05e55794fd7cb00R17-R22): Added a reference to the new post-load script `postload-mysql.sql` to create indices after loading the database, which improves performance by nearly 30%.

Codebase simplification:

* [`src/main/resources/benchmarks/tpch/ddl-mysql.sql`](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL19): Removed all index creation statements from the DDL script to speed up the initial load process. [[1]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL19) [[2]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL29-L30) [[3]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL44) [[4]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL57-L58) [[5]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL70-L73) [[6]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL87-L88) [[7]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL103-L105) [[8]](diffhunk://#diff-d0e0637fb2f0a9ee276e7046d3e142c830e9a652deb6a15afb2a1733e15d3caeL128-L135)
[...]

PR #596, opened 2025-02-13
Enable TPC-H for MariaDB in CI

### LINKED ISSUE #333: MariaDB TPC-H test takes a long time
Disabled via #330 for now, but we should dig into this and see why it's taking so much longer than other engines.  e.g., perhaps it's missing an index, has some extras, etc.

bpkroth:
Fixes #333

Following #594


bpkroth:
@nitinbil fyi

PR #597, opened 2025-02-13
Add some devcontainer fixups for running docker in docker

bpkroth:
Slight tweaks to allow running docker inside the devcontainer again.


bpkroth:
@eujing can you please try this out and let me know if it helps?  You'll probably have to rebuild your devcontainer without cache with this branch checked out.

eujing:
@bpkroth just tried this out, this definitely addressed the permission issues!
Now I can use docker from within the devcontainer, and perform the docker builds.

PR #600, opened 2025-02-13
Add ability to load after scripts from local filesystem

### LINKED ISSUE #595: The XML option `afterload` only works with packaged scripts
To speed up loading for the TPCH benchmark with MySQL, we tried overwriting DDL with `ddlpath` and `afterload` settings in our configuration file.
Essentially, `afterload` was pointing to a custom .sql script to create indices after loading was done.

Loading completes successfully with the custom DDL. However, we get a NullPointerException for the `afterload` step as it is trying to load the specificed script via `getResourceAsStream()`, but the script on the filesystem, not packaged with the .jar file.

Attached stack trace:
Image

https://github.com/cmu-db/benchbase/blob/46fc66f39de7f598de917220bbe12f1651943e3a/src/main/java/com/oltpbenchmark/util/ScriptRunner.java#L61

bpkroth:
Still fallsback to loading from resource stream when file is not found.

So far logging is only done to the debug stream.

Quick (hackish) fix to close #595

### DIFF
--- a/src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
+++ b/src/main/java/com/oltpbenchmark/api/BenchmarkModule.java
@@ -19,6 +19,7 @@ import com.oltpbenchmark.WorkloadConfiguration;
 import com.oltpbenchmark.catalog.AbstractCatalog;
 import com.oltpbenchmark.types.DatabaseType;
 import com.oltpbenchmark.util.ClassUtil;
+import com.oltpbenchmark.util.FileUtil;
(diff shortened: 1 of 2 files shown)


bpkroth:
@eujing wanna try this out too?

bpkroth:
note: this is basically what's done for the `ddlpath` override already

motus (approved):
Looks good!

eujing:
> @eujing wanna try this out too?

Can confirm, `afterload` is now working with a .sql script from the filesystem with these changes!

